PR #365, opened 2026-07-17
chore(release): prepare native v3.0.0

rizumita:
## Problem

The authoritative native Rust distribution is ready to replace the Python-era v2 release unit, but the release workflow, platform contract, installer atomicity, version metadata, and production promotion policy were not yet closed enough for a v3 release.

## Contract change

- Set the Rust workspace and VS Code extension to 3.0.0 and cut the changelog section.
- Assemble one complete 21-file release unit before a single tag-only publish step.
- Pin release actions, Rust 1.85.0, and vsce 3.9.2.
- Close the supported platform matrix to macOS arm64, Linux x64/arm64 on glibc 2.39+, and Windows x64.
- Stage and checksum both native binaries before installer replacement.
- Add the production source-policy gate and native Rust release contract coverage.
- Correct the generic annotation documentation to match the checked-model-only projection currently implemented.
- Add bounded browser-worker and native subprocess timeouts without retries.

This branch includes the latest main after #347 and #348 were conflict-resolved, revalidated, and merged.

## Evidence

- `./tools/check-native-integration.sh` — pass, including full Rust workspace, LSP, native/browser Z3, Worker cancellation, and 345 native/WASM parity cases.
- auto-engine regression: 15/15 pass.
- literate Markdown regression: 12/12 pass.
- native release integration contract: 6/6 pass.
- VSIX 3.0.0 packaged with Apache-2.0 license; npm audit reports 0 vulnerabilities.
- characterization baseline regenerated by its documented command and mechanically confirmed to differ only by 2.7.0 -> 3.0.0 values.

## Release note

This PR prepares but does not publish v3.0.0. After merge, the release workflow will be dispatched on the exact main SHA as a non-publishing dry run before production promotion and tagging.

## Linked issue

None.


rizumita (commented):
Retrospective audit (already merged, no action needed — recording findings for the record).

No material bugs found in the diff itself — `install.sh`'s staged-checksum-then-atomic-`mv` logic is sound, and version bumps (`Cargo.toml`/VS Code extension/`CHANGELOG.md`) are consistent.

Worth recording as a process lesson: the release-readiness evidence (15/15, 12/12, 6/6, `check-native-integration.sh`) never actually built against the pinned Rust 1.85.0 toolchain — the PR explicitly deferred that to a post-merge dry run, which then failed on all 4 targets (root cause: z3-sys build-deps needing 1.88, fixed in #368). The new `native_integration.rs` assertions (e.g. `workflow.contains("toolchain: 1.85.0")`) check workflow-file structure, not actual buildability. Per #370's follow-ups (missing `z3.dll` on Windows, Z3 archive 403 on macOS), at least 2 of the 4 "closed" platform-matrix targets weren't actually green at merge time either.

For future release-readiness PRs: building with the exact pinned toolchain before merge (rather than deferring to a post-merge dry run) would have caught this earlier. Also still open: `ci.yml` uses a floating `rust-toolchain@stable`, so ordinary CI still can't detect this class of drift going forward.

PR #368, opened 2026-07-17
fix(release): align Rust MSRV with locked dependencies

rizumita:
## Problem
The exact-main release dry-run failed on all four native targets because the pinned Rust 1.85.0 toolchain cannot build locked dependencies requiring Rust 1.88.

## Contract change
- Declare Rust 1.88 as the workspace MSRV.
- Pin release builds to Rust 1.88.0.
- Apply behavior-preserving let-chain consolidations required by the newly effective Clippy MSRV.

## Evidence
- `cargo +1.88.0 check --manifest-path rust/Cargo.toml --workspace --locked`
- `Z3_SYS_Z3_VERSION=4.16.0 cargo +1.88.0 build --manifest-path rust/Cargo.toml --release --locked -p fslc-rust -p fsl-lsp`
- `cargo clippy --manifest-path rust/Cargo.toml --workspace --all-targets --locked -- -D warnings`
- `./tools/check-native-integration.sh` (345 native/WASM parity cases)

## Release evidence
Failed non-publishing dry-run: https://github.com/ymm-oss/fsl/actions/runs/29551023546
No tag or GitHub Release was created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4
         with:
-          toolchain: 1.85.0
+          toolchain: 1.88.0
 
       - name: Build native Rust binary
         working-directory: rust
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ and versioning follows [Semantic Versioning](https://semver.org/). Each version 
   VSIX, and Kernel bundles before one tag-only publication step. Release actions,
   the Rust toolchain, and the VSIX packager are pinned; publication verifies a
   remote draft inventory before making it public, and manual dry runs validate
-  the same assembled unit without publishing it.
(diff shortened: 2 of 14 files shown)


rizumita (commented):
Retrospective audit (already merged, no action needed — recording findings for the record).

Manually verified all 8 let-chain consolidations are behavior-preserving (one non-obvious case in `rust/fsl-tools/src/analysis_graph.rs:328` double-checked against the invariant that `edge["kind"]` is always single-valued — confirmed equivalent). `cargo check`/`clippy` clean on current main. Pulled the actual failed Action log (run 29551023546) — root cause is `zip@8.6.0`/`time@0.3.53` (z3-sys build-deps) requiring 1.88; raising MSRV was a reasonable, conservative call (7 versions behind local 1.95).

One process gap worth tracking: no MSRV CI guard was added — `ci.yml`/`pages.yml` still float on `rust-toolchain@stable`, so a future `cargo update` could silently re-raise the effective MSRV again, undetected until the next release dry run (same gap noted on #365). Might be worth a follow-up issue for a dedicated `cargo +1.88.0 check` CI job.

PR #829, opened 2026-08-19
fix(docs-site): enforce ja/en LANGUAGE section order, not just count

### LINKED ISSUE #741: build_site_reference.py enforces section count only: a reordered LANGUAGE.ja.md silently mispairs bodies with English headings despite D7's "same count and order" claim
## Gap

`docs/DESIGN-docs-site.md` D7 states the two canonical language files are kept
section-aligned 1:1 — "the same count and order of `## ` headings" — and its
drift bullet claims the generation-time check "catches an added/removed/**reordered**
section". The docstring of `tools/build_site_reference.py` repeats the claim
("same count and order of \"## \" sections").

The mechanical check enforces **count only**. In `render_language_tree()` in
`tools/build_site_reference.py`:

- the ja branch raises `SystemExit` only when
  `len(render_sections) != len(en_sections)`;
- the two languages are then paired **positionally** with
  `zip(en_sections, render_sections)`, and the paired headings are never
  compared. The English heading of each pair supplies the anchor slug and the
  `SECTION_BLURBS` lookup; the Japanese heading and body are rendered under it
  unchecked.

## Failure mode

Reordering `## ` sections in `docs/LANGUAGE.ja.md` (e.g. swapping "## 2. 型"
and "## 3. 式") keeps the section count at 18, so the generator succeeds and
`docs/intro/language.ja.html` ships with Japanese section bodies attached to
the **wrong** English anchors and blurbs. `tests/test_site_reference_snapshot.py`
and the required `site reference freshness` context then *enforce* the misaligned
page rather than reject it: they only compare the committed HTML against a fresh
regeneration, so once the mispaired page is committed the gate is green.

This is an existing accepted invariant (D7) that is weaker in enforcement than
its stated contract — an added or removed section is caught, a pure reorder is
not.

## What a fix must assert

Heading **correspondence** for each `zip`-pair, not just equal counts. The
headings are translated, so byte equality cannot be the check, but 17 of the 18
sections in both files carry the same numeric prefix (`## 1.` … `## 17.`), and
the remaining lead section ("Design principles" / "設計原則") is unnumbered in
both. A minimal sound assertion: for every positional pair, the numeric section
prefix must be identical, and a numbered section must never pair with an
unnumbered one — failing with a `SystemExit` diagnostic that names both
headings and the position, in the same style as the existing count-mismatch
message. (An explicit ja↔en heading map would be equally sound; the prefix
check needs no new maintained table.)

## Rejecting fixture the fix must include

A calibrated negative control, per the repository's controls policy: feed the
checker a variant of `docs/LANGUAGE.ja.md` with two `## ` sections swapped
(count unchanged) and assert the generator exits nonzero with the
heading-correspondence diagnostic. An accepting control on the untouched files
must stay green. Without the rejecting fixture there is no evidence the new
check can detect the drift it exists for.

## Scope note

Found in passing during the #737 spike evaluation; it is independent of that
issue's outcome (the C1 changelog-fragment decision does not touch
`docs/LANGUAGE.md`, `docs/LANGUAGE.ja.md`, or any of this enforcement).

rizumita:
## Problem

`docs/DESIGN-docs-site.md` D7 states the two canonical language files are kept section-aligned 1:1 —
"the same count **and order** of `## ` headings" — and claims the generation-time check catches an
added, removed, or **reordered** section. `tools/build_site_reference.py`'s docstring repeated it.

**The check enforced count only.** `render_language_tree()` raised `SystemExit` when
`len(ja_sections) != len(en_sections)`, then paired the two languages **positionally** with
`zip(en_sections, render_sections)` and never compared the paired headings. The English heading of
each pair supplies the anchor slug and the `SECTION_BLURBS` lookup; the Japanese heading and body
were rendered under it unchecked.

Reproduced on `429b02f` before the fix: swapping the "2."/"3." sections in a copy of
`docs/LANGUAGE.ja.md` left the count at 18, `render_language_tree("ja")` **returned normally**, and
the output rendered anchor `id="2-types"` carrying the Japanese heading `3. 式` and its body.

**And the freshness gate cannot substitute for the missing check, by construction.**
`tests/test_site_reference_snapshot.py` and the required `site reference freshness` context compare
the *committed* HTML against a *fresh regeneration from the same sources*. That can only catch a
**stale** page. It structurally cannot catch a **wrongly generated** one: commit a reordered
`LANGUAGE.ja.md` alongside its mispaired HTML and the gate sees them agree and passes — enforcing
the drift instead of catching it. Order has to be asserted inside the generator.

## Contract change

A second check in the same function: for every positional pair, the heading's leading numeric prefix
must match (`"2"` from `"2. Types"` / `"2. 型"`; `None` for the unnumbered lead section), raising
`SystemExit` naming both headings and the 1-based position, phrased as a sibling of the existing
count-mismatch message.

This needs **no new maintained ja↔en heading table**, because both files already share one numbering
scheme — measured: both have 18 `## ` sections, an unnumbered lead ("Design principles" / "設計原則")
followed by `## 1.` … `## 17.` in identical order. The residual limitation is recorded in D7: a
reorder among *unnumbered* sections would not be caught, but only the lead section lacks a prefix
today, so no such reorder is currently possible.

## Test evidence

- **Rejecting control** — `test_language_reference_rejects_reordered_ja_sections` builds the swapped
  variant in `tmp_path` from the real file's content at test time and points `LANGUAGE_MD_JA` at it
  via `monkeypatch`; the committed `docs/LANGUAGE.ja.md` is never written. It asserts
  `len(swapped) == len(sections)` inline, so the test itself records that the count check alone
  cannot catch this.
- **Accepting control** — `test_language_reference_accepts_aligned_real_files` generates both trees
  from the untouched real files, so the fix is shown not to break the normal path.

**Calibration verified by mutation**: removing only the new correspondence loop makes the rejecting
control fail with `DID NOT RAISE SystemExit` while the other four tests stay green — so that control
is the only mechanized detector of this regression, and it does detect it. Restored afterwards;
5/5 pass.

These controls are machine-enforced, not dormant: `site-reference-freshness.yml:77` runs
`pytest tests/test_site_reference_snapshot.py` — the whole file — and that context is a required
status check in the `main safety and CI` ruleset, on every pull request with no path filter.

`python3 tools/build_site_reference.py` succeeds and `docs/intro/` is byte-identical: the change adds
a check without altering how correctly-aligned files render.

## Documentation

`docs/DESIGN-docs-site.md` D7's two bullets that named count comparison as the mechanism are
corrected, plus two addenda: one recording the gap and its fix with the residual limitation, one
recording the freshness gate's structural inability to substitute for it. `docs/LANGUAGE.md` and
`docs/LANGUAGE.ja.md` are untouched — they are correctly aligned today, and editing a canonical
source to satisfy a check would be backwards.

`changelog.d/741-language-reference-section-order.fixed.md`.

Closes #741.

### DIFF
--- a/changelog.d/741-language-reference-section-order.fixed.md
+++ b/changelog.d/741-language-reference-section-order.fixed.md
@@ -6,6 +6,9 @@ attached a Japanese section body to the wrong English anchor and blurb — exact
 `docs/DESIGN-docs-site.md` D7 and this tool's own docstring already claimed was caught. The
 fix checks each positional pair's leading numeric section prefix (`"2"` from `"2. Types"`/
 `"2. 型"`) and raises `SystemExit` naming both headings and the position on a mismatch, needing
-no new maintained ja/en heading-name table. A rejecting control (a same-count reordering
-fixture) and an accepting control (the real, untouched files) are both in
+no new maintained ja/en heading-name table. That per-position check is only sound if
+`docs/LANGUAGE.md`'s own numbers are unique, so a second assertion now rejects a duplicated
+English section number directly, naming the duplicate, instead of leaving uniqueness an
+unstated precondition. Rejecting controls (a same-count reordering fixture and a duplicate-number
+fixture) and an accepting control (the real, untouched files) are all in
 `tests/test_site_reference_snapshot.py`.
--- a/changelog.d/741-language-reference-section-order.fixed.md
+++ b/changelog.d/741-language-reference-section-order.fixed.md
@@ -1,0 +1,11 @@
+Fixed (#741): `tools/build_site_reference.py` now rejects a `docs/LANGUAGE.ja.md` that
+reorders `## ` sections relative to `docs/LANGUAGE.md` while keeping the section count equal.
+`render_language_tree()` previously compared only `len(ja_sections) == len(en_sections)` and
+then paired the two files positionally, so a same-count reorder passed generation silently and
+attached a Japanese section body to the wrong English anchor and blurb — exactly the drift
+`docs/DESIGN-docs-site.md` D7 and this tool's own docstring already claimed was caught. The
+fix checks each positional pair's leading numeric section prefix (`"2"` from `"2. Types"`/
+`"2. 型"`) and raises `SystemExit` naming both headings and the position on a mismatch, needing
+no new maintained ja/en heading-name table. A rejecting control (a same-count reordering
+fixture) and an accepting control (the real, untouched files) are both in
+`tests/test_site_reference_snapshot.py`.
--- a/docs/DESIGN-docs-site.md
+++ b/docs/DESIGN-docs-site.md
@@ -187,11 +187,13 @@ readable in Japanese, accepting the resulting maintenance cost.
   keys/values, and every fenced code block are left untranslated/byte-identical to the English
   source — only heading text and prose are translated.
 - **How the site stays honest about drift:** `render_language_tree("ja")` in
-  `tools/build_site_reference.py` reads `docs/LANGUAGE.ja.md` and compares its section count
-  against `docs/LANGUAGE.md`'s. A mismatch raises `SystemExit` at generation time rather than
-  silently shipping a page with a missing or misplaced section — the same "fail loudly on drift"
-  discipline D4/the original docstring already used for `SECTION_BLURBS` coverage, now also
-  covering section-count parity between the two language files.
+  `tools/build_site_reference.py` reads `docs/LANGUAGE.ja.md`, compares its section count against
+  `docs/LANGUAGE.md`'s, and then checks that every positional pair of headings actually
+  corresponds by numeric prefix (see Addendum (#741) below — count parity alone does not imply
+  order parity). A mismatch on either check raises `SystemExit` at generation time rather than
+  silently shipping a page with a missing, misplaced, or reordered section — the same "fail loudly
+  on drift" discipline D4/the original docstring already used for `SECTION_BLURBS` coverage, now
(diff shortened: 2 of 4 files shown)


rizumita:
## Reviewer-demonstrated bypass, re-measured after the fix

The independent review found that the numeric-prefix comparison was `pairwise` only, and demonstrated a bypass: with both files numbered `1., 2., 2., 4.` and the two duplicated sections swapped in `docs/LANGUAGE.ja.md`, generation **passed** — the exact mispairing this change exists to reject.

The revision adds a uniqueness assertion on `docs/LANGUAGE.md`'s section numbers. I reproduced the reviewer's scenario against the revised generator to confirm it is closed:

```
en prefixes after renumbering 3. -> 2.:  ['Design principles', '1', '2', '2', '4', '5']
ja: the two duplicated sections swapped against their English counterparts
render_language_tree("ja")  ->  SystemExit
  "docs/LANGUAGE.md has more than one '## ' section numbered '2': ['2. Types', '2. Expressions']"
```

The assertion fires **before** any positional pairing, so that input cannot reach the correspondence loop at all — the bypass is closed by construction rather than by a wider comparison. The dedicated rejecting control (`test_language_reference_rejects_duplicate_en_section_numbers`) pins the assertion itself: removing only the uniqueness block makes exactly that test fail while the other five stay green.

Why this mattered more than its severity: this pull request exists because D7 claimed an enforcement it did not have, and the first revision's addendum recorded only the *unnumbered*-section limitation while leaving the uniqueness assumption unstated. Documenting the assumption would have been the same defect at smaller scale, so it was closed with an assertion instead — an assertion cannot be misread.

Issue #786, opened 2026-08-08
test: tests/test_dialect_conformance.py has 2 undetected failures (harness cited as a gate is red, unnoticed since CI never runs it)


rizumita:
## Problem

`tests/test_dialect_conformance.py` is the harness `AGENTS.md`/`CONTRIBUTING.md`/`docs/DESIGN-conformance-harness.md` cite when telling contributors to register a new dialect construct in `tests/dialect_registry.py`. It is not run by any CI workflow (`.github/workflows/`), not by `tools/check-native-integration.sh` (0 occurrences of `python`/`pytest` in that script), and not by `tools/check-merge-readiness.sh` either (that script directly execs two unrelated `tests/*.py` functions — `test_codex_environment`/`test_claude_environment` — without pytest, but never this file). Because of that, it has been silently red on 2 of 217 collected cases for some time, with nobody noticing.

Reproduced at `b470535` (current `main`):

```
PYTHONPATH=src <repo>/.venv/bin/python -m pytest tests/test_dialect_conformance.py -q
...
FAILED tests/test_dialect_conformance.py::test_full_pipeline[examples/e2e/2_requirements.fsl:CONFORMANCE]
FAILED tests/test_dialect_conformance.py::test_refinement_mapping_parses[examples/layers/return_impl_refines.fsl:REFINEMENT]
2 failed, 215 deselected in 1.24s
```

This report intentionally only *records* the two failures (with a defect-vs-staleness classification for each) rather than fixing them; a companion PR fixes the wording in `AGENTS.md`/`CONTRIBUTING.md`/the design doc/the two test docstrings so they stop describing this harness as a machine-enforced CI gate.

## Finding 1 — `test_full_pipeline[examples/e2e/2_requirements.fsl:CONFORMANCE]`

```
result = {'fsl': '1.0', 'result': 'error', 'kind': 'type',
          'message': "duplicate action map for 'submit'", 'loc': {'line': 38, 'column': 5}}
```

`examples/e2e/2_requirements.fsl` declares an explicit action-correspondence override for `submit` inside its `implements ExpenseToBe from "1_business.fsl" { action submit(c, a) -> submit(c) }` block, because `submit`'s business-layer arity (`(c)`) differs from its requirements-layer arity (`(c, a)`), so `maps auto` cannot infer it (see the file's own comment, and issue #494).

[...]

PR #837, opened 2026-08-19
test(conformance): pin the native-only refinement parser boundary by exact path and location

rizumita:
## Problem

`tests/test_dialect_conformance.py::test_refinement_mapping_parses` failed on
`examples/layers/return_impl_refines.fsl`:

```
lark.exceptions.UnexpectedCharacters: No terminal matches 'e' ... at line 11 col 3
  enum abstraction case_state DSt -> SSt {
```

That file uses the `enum abstraction { }` mapping form, which
`docs/DESIGN-enum-member-identity.md:13`-`:16`, `:38`-`:52` introduced as **Rust-native-only**, with an
explicit decision that the frozen Python reference does not move. So the parse failure is expected
native-only syntax, not a Python defect — but the harness classified every `refinement` file identically
and called the frozen parser unconditionally, so it had no way to express that.

## Scope

**Finding 2 of #786 only.** `src/fslc/` is untouched — Finding 1 (`duplicate action map for 'submit'`,
a divergence inside the frozen reference) is a separate decision and **still fails after this change**,
by design. The harness's CI status is also unchanged: it is a developer-run manual check, and
`AGENTS.md:92`-`:96` already says so accurately, so no wording was edited.

## Contract change

A typed, **exact-path** registry in `tests/dialect_registry.py`:

```python
NATIVE_ONLY_REFINEMENT_SYNTAX["examples/layers/return_impl_refines.fsl"] = NativeOnlyRefinementSyntax(
    path=..., construct="enum abstraction", line=11, column=3,
    design_citation="docs/DESIGN-enum-member-identity.md:13-16, 38-52",
    reason="source-total many-to-one enum abstraction mapping syntax is Rust-native-only; "
           "the accepted decision explicitly retains the frozen Python reference unchanged",
    native_owner="rust/fsl-core refinement typechecking and lowering",
)
```

A new `NATIVE_ONLY_REFINEMENT` classification is checked **before** the generic `REFINEMENT` case, and
`classify()` is split so `_classify_source(path, src, rel)` can be driven with synthetic inputs — that is
how the rejecting controls are built without editing any corpus file.

Deliberately **not** `MONITOR_EXCLUSIONS`: those assert Monitor non-loadability via `can_monitor`
(`tests/test_dialect_conformance.py:200`-`:216`), and a mapping file is non-monitorable whether or not
its syntax parses. The registry records that distinction in a comment.

## Why the location is load-bearing — this is what keeps it from being an allowlist

The test asserts the registered construct still exists **at the recorded line/column**, then requires the
parse to fail **exactly there**. Three distinct outcomes:

| Observed | Meaning | Test |
|---|---|---|
| parse succeeds | the exclusion is stale | **fails**: "frozen parser now accepts native-only …; remove the stale exclusion" |
| fails at the declared location | native-only by design | passes |
| fails elsewhere | a real Python parser regression | **fails**: "parser failure moved from … to …" |

Most exclusion mechanisms have only the middle row. Without the first, the entry outlives its cause and
nobody can tell; without the third, it swallows a genuine regression in that file. All five registry
fields are asserted non-empty, so an entry cannot be added without its justification.

Registration is by exact path, not by detecting `enum abstraction` anywhere — a new corpus file using
native-only mapping syntax **must** fail in the generic `REFINEMENT` case until someone reviews and
registers it.

## Controls

- **Accepting** — `test_native_only_refinement_exclusion_preserves_ordinary_mapping_parse`: an ordinary
  Python-compatible refinement still classifies as `REFINEMENT` and parses.
- **Rejecting (wrong location)** — `test_native_only_refinement_wrong_location_is_a_regression`: injects
  an ordinary syntax error *before* the declared token and requires the "parser failure moved" assertion
  to fire.
- **Rejecting (shape is not enough)** — `test_native_only_refinement_construct_is_not_excluded_by_shape`:
  an **unregistered** copy of the same source stays in generic `REFINEMENT` and fails to parse.

## Measured, independently of the implementer

```
before   217 collected   215 passed   2 failed
after    220 collected   219 passed   1 failed
```

The failure that disappeared is exactly `test_refinement_mapping_parses[…return_impl_refines.fsl…]`.
The one that remains is `test_full_pipeline[examples/e2e/2_requirements.fsl:CONFORMANCE]` with
`duplicate action map for 'submit'` — Finding 1, out of scope, and **it must still fail**: had it started
passing, that would be evidence of an unintended side effect rather than progress.

`changelog.d/786-native-refinement-exclusion.fixed.md`. No corpus file, no `src/fslc/`, no CI wiring, and
no design-document wording was changed.

Partially addresses #786 — Finding 1 stays open.

### DIFF
--- a/changelog.d/786-native-refinement-exclusion.fixed.md
+++ b/changelog.d/786-native-refinement-exclusion.fixed.md
@@ -1,1 +1,1 @@
-Fixed (#786): Classify the native-only `enum abstraction` refinement mapping with an exact-path, exact-location parser exclusion.
+Fixed (#786): Prove the native-only `enum abstraction` refinement exclusion by neutralizing only its registered block and reject stale or shadowed registry entries.
--- a/changelog.d/786-native-refinement-exclusion.fixed.md
+++ b/changelog.d/786-native-refinement-exclusion.fixed.md
@@ -1,0 +1,1 @@
+Fixed (#786): Classify the native-only `enum abstraction` refinement mapping with an exact-path, exact-location parser exclusion.
--- a/docs/DESIGN-conformance-harness.md
+++ b/docs/DESIGN-conformance-harness.md
@@ -86,6 +86,13 @@ exhaustively; the registry says what may exist there.
   individual files the frozen Python Monitor legitimately rejects. Each entry
   names its active native or BMC-side coverage, and a stale entry fails the
   harness once the Monitor starts accepting it.
+- `NATIVE_ONLY_REFINEMENT_SYNTAX: dict[str, NativeOnlyRefinementSyntax]` —
+  repo-relative exact path → native-only refinement parser boundary. Each entry
+  records the construct text and source location plus the accepted design and
+  native owner. The registry is mutually exclusive with `MONITOR_EXCLUSIONS`
+  and every earlier whole-file classification: every entry must exist on disk
+  and classify as `NATIVE_ONLY_REFINEMENT`, rather than silently winning or
+  losing by classifier order.
 
 ## Classification (automatic, in the harness)
 
@@ -93,15 +100,20 @@ exhaustively; the registry says what may exist there.
 
 1. `EXCLUDED` — `is_ai_project_source` / `is_ai_agent_source` / `is_causal_source`
    match, or path in `MONITOR_EXCLUSIONS`.
-2. `REFINEMENT` — top-level keyword `refinement` (mapping files are not state
+2. `NATIVE_ONLY_REFINEMENT` — an exact registered refinement path whose frozen
+   parser rejects a native-only construct. The harness confirms the parser's
+   diagnostic location, then removes only that registered block construct and
+   requires the rest of the source to parse; success proves the native syntax,
+   rather than another deferred error, is the sole exclusion reason.
+3. `REFINEMENT` — top-level keyword `refinement` (mapping files are not state
    machines; refine semantics are covered by `test_refine*.py` and the
    refinement fixtures in `test_oracle_agreement.py`).
-3. `DECLARED_ERROR` — front matter `// expected-result: error` (gallery error /
+4. `DECLARED_ERROR` — front matter `// expected-result: error` (gallery error /
    adversarial fixtures that must fail at parse/type/semantics/acceptance).
-4. `INJECTED` — `// inject:` / `// expect-detector:` front matter
+5. `INJECTED` — `// inject:` / `// expect-detector:` front matter
    (`examples/gallery/injected/`, the detector benchmark corpus of
    `test_injection_bench.py`).
-5. `CONFORMANCE` — everything else; the top-level keyword must match a
+6. `CONFORMANCE` — everything else; the top-level keyword must match a
    `DIALECTS` construct. **An unknown construct fails the run** with
    "register the new dialect in tests/dialect_registry.py".
 
@@ -132,6 +144,7 @@ One parametrized test per class; every obligation is an `assert`, never a skip.
 |---|---|
 | CONFORMANCE / INJECTED | full pipeline below |
 | REFINEMENT | `parse_src` succeeds and `ast[0] == "refinement"` |
+| NATIVE_ONLY_REFINEMENT | frozen `parse_src` still rejects the registered construct, and removing only that construct's matched-brace span lets the remainder parse; any other error makes the exclusion invalid |
 | DECLARED_ERROR | build or `run_verify` still yields `result:"error"` — a fixture that starts passing is a stale declaration and fails |
 | EXCLUDED | the documented reason still holds (Monitor load still raises / construct still matches) — a stale exclusion fails and must be deleted |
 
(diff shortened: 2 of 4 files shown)


rizumita:
## Review found a merge blocker: the location check alone accepts a real regression

Independent review, then reproduced by me. **This PR should not merge as it stands.**

`_assert_native_only_refinement_failure` compares only `(line, column)`. That is necessary but **not
sufficient**, because a different parse error can surface at the same coordinates.

Reproduced independently: replace line 3, 4 **or** 5 of `examples/layers/return_impl_refines.fsl` with
`"  preserve progress {"` — a genuine syntax corruption, line count preserved — and:

```
original                       fails at 11:3
line 3 replaced                fails at 11:3
line 4 replaced                fails at 11:3
line 5 replaced                fails at 11:3
```

Lark defers detection because the unterminated block swallows the intervening comment lines until it
reaches the native construct. So the assertion passes and the corruption is silently accepted. The third
advertised outcome — "failure anywhere else is a regression" — does not hold. The existing wrong-location
control only covers an error reported *immediately* earlier, which is exactly why this got through.

The fix is not a second coordinate check. It is a decisive property: **neutralize only the registered
construct's span and require the rest of the file to parse successfully.** That proves the native syntax is
the *sole* reason the file fails, which is the claim the whole exclusion rests on. Comparing the expected-token
set is a reasonable secondary signal (the original expects top-level refinement tokens; the mutated one
expects `RESPOND`/`RBRACE`), but token sets are a Lark implementation detail and pinning them alone would make
the test brittle across parser versions.

## Two more, both confirmed

**The registry is not exhaustively validated.** `tests/test_dialect_conformance.py:321` validates every
`MONITOR_EXCLUSIONS` path; nothing does the equivalent for `NATIVE_ONLY_REFINEMENT_SYNTAX`, and the controls
inspect only `NATIVE_ONLY_REFINEMENT_CASES[0]`. Demonstrated: a nonexistent second entry passes quietly, and
an entry that also appears in `MONITOR_EXCLUSIONS` classifies as `EXCLUDED` because that check precedes
native-only classification at `:91` — silently shadowed. Every entry needs: path exists, classifies as
`NATIVE_ONLY_REFINEMENT`, not shadowed. Whether an overlap is a hard error or an ordering change is a
deliberate choice that should be stated, not resolved by quietly reordering.

**The accepted design document is stale.** `docs/DESIGN-conformance-harness.md`'s "Registry" section documents
only `MONITOR_EXCLUSIONS`, and its "Classification (automatic, in the harness)" section enumerates what
`classify()` returns without `NATIVE_ONLY_REFINEMENT`, so its obligation table omits it. `AGENTS.md`'s
coupled-change rule requires the design to move with the harness.

## What still holds

The registry shape and its five justification fields; the exact-path design; the "parse succeeded means the
exclusion is stale" outcome; the accepting control; the no-op-replace, compatible-shape and empty-list attacks
(all fail loudly — empty lists raise `IndexError`); line insertion and deletion above the construct (both fail
loudly); classification comparison across the real corpus showing only the intended file changed; the diff
footprint (`examples/`, `specs/` and `src/fslc/` byte-identical); SPDX; four-space Python; and the
`Fixed (#786):` fragment prefix. The measured harness result — 220 collected, 219 passed, 1 failed, with the
out-of-scope Finding 1 case still failing as required — also holds.

The reviewer noted its own coverage limits: targeted pytest only (8 passed, 216 deselected), no full harness
run, no Cargo commands, and it relied on my measurement for the continued Finding 1 failure.

rizumita:
## All three findings fixed, and the blocker's closure verified independently

`cc33871` on top of the previous two commits.

### Finding 1 — the deferred-error hole is closed

The blocker was that comparing only `(line, column)` accepted a real regression: replacing line 3, 4 or 5 of
`examples/layers/return_impl_refines.fsl` with `"  preserve progress {"` still failed at exactly `11:3`,
because Lark defers detection through the unterminated block.

Re-ran that exact reproduction against the fixed assertion. **All three cases are now rejected:**

```
line 3 replaced  ->  rejected: "neutralizing only registered native-only construct ..."
line 4 replaced  ->  rejected: same
line 5 replaced  ->  rejected: same
```

The fix replaces the coordinate comparison with the decisive property: neutralize **only** the registered
construct's span and require the remainder to parse successfully. Coordinates can coincide; "this construct is
the sole reason the file fails" cannot be faked by an unrelated error landing at the same place. That is what
the exclusion actually rests on, so it is what the test now asserts.

### Findings 2 and 3

Every registry entry is now validated — path exists, classifies as `NATIVE_ONLY_REFINEMENT`, and is not
shadowed by an earlier classification — instead of only inspecting `NATIVE_ONLY_REFINEMENT_CASES[0]`. And
`docs/DESIGN-conformance-harness.md` gains the new registry and classification, describing what the
classification asserts rather than merely naming a label, so the accepted design states the real contract.

### Measured

```
original main            217 collected   215 passed   2 failed
this PR, first version   220 collected   219 passed   1 failed
this PR, after fixes     223 collected   222 passed   1 failed
```

Three controls added by the fixes. The single remaining failure is
`test_full_pipeline[examples/e2e/2_requirements.fsl:CONFORMANCE]` — `duplicate action map for 'submit'`, which
is **Finding 1 of #786**, out of scope here, and pending a separate decision about changing the frozen
`src/fslc/` reference. **It must still fail**, and it does: had it started passing, that would have been
evidence of an unintended side effect rather than progress.

No corpus file under `examples/` or `specs/` was touched, and `src/fslc/` is unchanged.

Issue #795, opened 2026-08-12
meta(ci): a failing cache budget audit files no issue, so #747's recurrence ran unobserved for four days


rizumita:
## Problem

`cache budget audit` failed on `main` for four consecutive days (2026-08-08 through 08-11)
with `budget-exhausted: cache usage is 10.03 GiB of a 10.00 GiB limit (100%)`, and **no issue
tracked it**. #747 — the same defect class — had been closed on 2026-08-06.

The budget itself is fixed by #794. This issue is about the two process gaps that let the
recurrence run unobserved for four days:

### 1. A failing audit produces no tracked work item

`cache-budget-audit.yml` reports and exits non-zero, but nothing files or updates an issue.
`post-merge-ci-reporter.yml` does exactly that for the product gate (see #617, and the
`[post-merge CI]` issues closed on 2026-08-06/08). The cache audit has no equivalent, so a
red audit is only visible to whoever happens to open the Actions tab.

Four consecutive failures produced zero tracked work.

### 2. The audit could not see the entries that actually disappeared

Found while fixing #794 and now fixed there, but worth recording as the reason the first gap
mattered so much: `sharedKeyOf`'s regex matched `Linux|macOS|Windows`, while `os.type()`
actually produces `Linux`, `Darwin`, and `Windows_NT`. Both `rust-native-z3` entries were
therefore invisible to key-level diagnosis — including the Windows entry whose eviction
started the incident. The audit could report `budget-exhausted` (a usage total, no key
parsing needed) but could never report that the load-bearing entry was gone.

So during those four days the audit was both **unheard** (gap 1) and **partially blind**
(gap 2).

## Why this matters

`AGENTS.md` requires that a post-merge product-gate failure is not hidden and that its
automatically created issue is not treated as a waiver. The cache budget audit is not the
product gate, but it guards the resource the product gate depends on: this incident's chain
was cache eviction → cold Windows build → 40-minute timeout → cancelled job → skipped cache
save → the same cold build next time. `main` had no complete product evidence from 2026-08-07
to 2026-08-12 as a result.

## Suggested scope

- Give the cache budget audit a reporting path comparable to `post-merge-ci-reporter.yml`:
  open an issue on failure, keep it open until a later run proves recovery, and avoid opening
  duplicates for the same ongoing failure.
- Decide explicitly whether the audit should be a required context. It is currently advisory,
  which is defensible — but that decision is not recorded anywhere, and the four-day silence
[...]

PR #839, opened 2026-08-20
test(ci): give the cache budget audit's workflow wiring a calibrated semantic validator

### LINKED ISSUE #802: meta(ci): cache budget audit の workflow 配線に回帰対照が無く、|| true / permissions / trigger の変更で監査を黙って無効化できる
# Problem

`.github/workflows/cache-budget-audit.yml` の配線に**回帰対照が一つも無い**。
runner 側（`.github/scripts/run-cache-budget-audit.mjs`）と純関数側
（`audit-cache-budget.mjs`）は 53 の rejecting control で厚く校正されているのに、
**その結果を CI の成否に変換する層は、壊しても全テストが緑のまま**である。

AGENTS.md:

> A confidently green false negative is more dangerous than a crash.

# 実測（2026-08-13、PR #797 のレビュー中に独立再現）

`node --test .github/scripts/audit-cache-budget.test.mjs`（53 tests）と `actionlint` を
基準に、workflow へ次の変異を当てた結果:

| 変異 | 結果 |
|---|---|
| `run: node .github/scripts/run-cache-budget-audit.mjs` → `... \|\| true` | **53 tests + actionlint 緑** |
| `permissions: actions: read` → `actions: none` | **緑** |
| `on.schedule` を削除 | **緑** |
| `on.push.branches: [main]` → `[never]` | **緑** |

つまり:

- **runner が exit 1 を返しても job を成功させられる**（`|| true` / `continue-on-error`）
- **live API を読めなくする権限変更が検出されない**
- **定期監査を黙って停止できる**
- **checker 変更後の push 監査を黙って停止できる**

いずれも「監査は動いているように見えるが、実際には何も守っていない」状態を作れる。

# 背景

#747 の cache-eviction / cold-build ループは、**監査が 4 日連続で失敗していたのに
誰も気づかなかった**という経緯を持つ（#795）。
本 issue はその隣接面で、**監査そのものが黙って無効化されうる**ことを指す。

PR #797 のレビューでは、runner 内部の false PASS 経路が
round-10 から round-14 まで繰り返し発見された。それらはすべて修正されたが、
**その修正が CI に反映される配線には対照が無い**ままである。

# 期待される挙動

`cache-budget-audit.yml` の次の性質を、**静的契約テストで固定**すること:

1. **audit step の終了状態が job の成否に直接伝播する。**
   `continue-on-error`、`|| true`、`set +e`、`if: always()` による握りつぶしを拒否する。
2. **`permissions` が `actions: read` と `contents: read` を保持する。**
3. **trigger が `schedule` / `workflow_dispatch` / `push`（`main`）を保持する。**
4. **監査関連の path 一覧**（runner / 純関数 / テスト / workflow 自身）が
   push trigger の対象に含まれている。

実装形態は問わない（Node のテストから YAML を読んで assert する、
`actionlint` のカスタムルール、専用の contract test など）。
**重要なのは、上記のいずれかを壊すと必ず落ちること。**

# 影響範囲

- `.github/workflows/cache-budget-audit.yml`
- 監査が守っている対象そのもの（#747 の再発検出）

# 関連

- **#795** — 監査が失敗しても issue が起票されず、#747 の再発が 4 日間観測されなかった。
  **本 issue は「監査が失敗を報告する経路」ではなく「監査が動作し job を落とす経路」を対象とする**
  ため別 issue とする。両方を同じ workflow で解くことは可能。
- **#747** — cache-eviction / cold-build ループ。
- **PR #797** — runner と純関数側の false PASS 経路を修正。
  そのレビュー（round-14）で本 issue の欠陥が発見された。
  **#797 は `cache-budget-audit.yml` を変更していない**ため、本 issue はその範囲外。

# 補足

同種の穴が他の workflow にも無いか確認する価値がある。
特に `.github/workflows/ci.yml` の product gate と
`.github/workflows/merge-readiness.yml` について、
「step の失敗が job の失敗に伝播すること」の静的対照があるかは未調査。

rizumita:
## Problem, corrected

The issue body says the cache budget audit's wiring has **no** regression control. Measured on `9754590`,
that overstates it:

- `node --test .github/scripts/audit-cache-budget.test.mjs` passes **65/65**, and
  `tools/check-merge-readiness.sh`'s `check_automation()` lane runs it on **every pull request** — with a
  comment naming its rejecting fixture (removing `ci.yml`'s `save-if` guard must fail the audit).
- Those tests execute the real CLI `main()`/`process.exit` path in a subprocess and prove that an unhealthy
  result and missing credentials both return status 1.
- The workflow itself currently has **zero** `continue-on-error`, `|| true`, `set +e` and `if:` across its 68
  lines, and declares exactly `actions: read` and `contents: read`.

The real gap is narrow and precise: **no test reads `.github/workflows/cache-budget-audit.yml`.** Runner
failure propagates today, but a future edit to that YAML — its triggers, its permissions, or the audit step's
invocation — is rejected by nothing. That is what this PR closes. The correction is recorded on the issue.

One criterion in the issue is also technically wrong and was **not** implemented as written: `if: always()`
does not turn a failed step green — it controls whether a step *runs* after a prior failure, while its exit
status still fails the job. The validator forbids **any** `if` on the audit step so it stays unconditional,
which is a defensible rule, but the rationale is "keep it unconditional", not "always() swallows exit status".

## What it does

`.github/scripts/validate-cache-budget-audit-workflow.py` (200 lines) validates a **parsed YAML document**,
locating the audit job and the named live-audit step semantically, then requiring:

| | |
|---|---|
| triggers | `schedule`, `workflow_dispatch`, `push` all present and non-empty; `push.branches` contains `main`; all six watched paths present |
| permissions | exactly the two read grants |
| ordering | calibration before the live audit |
| tolerance | no `if` and no `continue-on-error` on the step; no job-level tolerance |
| command | **allowlisted** as the single argv-equivalent `node .github/scripts/run-cache-budget-audit.mjs` |

The command is allowlisted rather than substring-scanned. A scalar containing a pipeline, a compound command,
`set +e`, a redirection wrapper, or any extra command is not that command. That makes quoting, indentation,
key order, comments and block-versus-plain YAML irrelevant, so the validator only needs editing when the
protected semantics actually change — the difference between a control that survives and one that gets edited
into uselessness.

## Calibration — the acceptance criterion

`tests/test_cache_budget_audit_workflow.py` parses the real workflow, deep-copies the semantic tree, and
applies **one mutation at a time**, requiring rejection for each. **16 rejecting cases plus 1 accepting case,
17/17 pass.**

The per-item mutations are **derived from the validator's own constants**, not hand-listed:

```python
*[(f"remove permission {p}", _remove_permission(p)) for p in sorted(validator.REQUIRED_PERMISSIONS)],
*[(f"remove trigger {t}",    _remove_trigger(t))    for t in validator.REQUIRED_TRIGGERS],
*[(f"remove watched path {p}", _remove_watched_path(p)) for p in sorted(validator.REQUIRED_WATCHED_PATHS)],
```

So widening what the validator protects automatically widens what the mutation table attacks. A hand-written
list would let the next person add a protected key and forget its mutation.

**Verified independently by the orchestrator, not taken from the implementer's report:** deleting only the
live-step `continue-on-error` check from the validator makes exactly
`test_each_single_wiring_mutation_is_rejected[set live-step continue-on-error]` fail — 1 failed, 16 passed —
and restoring it returns 17/17 with a clean tree. Each mutation is pinned individually, not by a single
umbrella assertion.

## Placement, and the dependency question

A real YAML parser is the right tool, so this cannot join `tools/check-merge-readiness.sh`'s automation lane:
that lane's own comment (`:36`-`:41`) states it stays stdlib-only with no pytest and no third-party
dependency, and that purity — not its speed — is what its placement protects.

It follows `site-reference-freshness.yml`'s precedent instead: third-party-dependent evidence gets its own
unfiltered `pull_request`/`merge_group` workflow. The install step defers to `pyproject.toml`'s unpinned dev
extra, exactly as that workflow does, because deferring to the declaration is what stops a hand-maintained
package list from going stale.

PyYAML is added to that dev extra. **This is not a change to the frozen Python product surface**: `AGENTS.md`
freezes `src/fslc/`'s behavior, and `markdown>=3` is already in the same extra for precisely this reason — it
is a third-party dependency of a repository tooling script (`tools/build_site_reference.py`), not of the
package.

## Deliberately out of scope

**The new context is not made required.** `.github/ruleset-contract.json` and `docs/DESIGN-ci.md`'s
required-context list are untouched. Making a context required is a coordinated rollout across the checked-in
contract, its captured fixtures, the design record and the live ruleset; bundling it here would make this
change hard to revert. The workflow is unfiltered so that rollout can be evaluated later without a
path-filter reporting gap, and its header says so.

The audit's inability to distinguish reclaimable superseded duplicates from genuine growth is filed
separately as #835. Failure-time issue creation is #795, which should extend this validator rather than
introduce a second wiring check.

`changelog.d/802-cache-budget-audit-wiring.required.md`.

Closes #802.

### DIFF
--- a/.github/scripts/validate-cache-budget-audit-workflow.py
+++ b/.github/scripts/validate-cache-budget-audit-workflow.py
@@ -165,10 +165,10 @@ def validate_workflow(document: Any) -> list[str]:
     if calibration_indices and calibration_indices[0] >= live_audit_index:
         errors.append("calibration step must precede the live-audit step")
 
+    # _named_step_indices selects only Mapping instances, so the unique live
+    # step above is necessarily a mapping.  Keeping a defensive non-mapping
+    # branch here would create an unreachable, uncalibrated diagnostic.
     live_step = steps[live_audit_index]
-    if not isinstance(live_step, Mapping):
-        errors.append("named live-audit step must be a mapping")
-        return errors
     if "if" in live_step:
         errors.append("live-audit step must not declare 'if'")
     if "continue-on-error" in live_step:
@@ -187,7 +187,16 @@ def main() -> int:
     parser.add_argument("--workflow", type=Path, default=DEFAULT_WORKFLOW)
     args = parser.parse_args()
 
-    errors = validate_workflow(load_workflow(args.workflow))
+    try:
+        document = load_workflow(args.workflow)
+    except FileNotFoundError:
+        errors = [f"workflow file '{args.workflow}' does not exist"]
+    except OSError as error:
+        errors = [f"workflow file '{args.workflow}' could not be read: {error.strerror}"]
+    except yaml.YAMLError as error:
+        errors = [f"workflow YAML is invalid: {error}"]
+    else:
+        errors = validate_workflow(document)
     if errors:
         for error in errors:
             print(f"cache-budget-audit workflow wiring: FAIL -- {error}")
--- a/.github/scripts/validate-cache-budget-audit-workflow.py
+++ b/.github/scripts/validate-cache-budget-audit-workflow.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Validate the semantic wiring of the Actions cache-budget audit workflow.
+
+The workflow is parsed before it is inspected so formatting, comments, key
+order, quoting, and plain-versus-block scalar style do not affect this
+contract.  The matching calibration suite derives each rejecting fixture from
+the live workflow rather than carrying a second workflow snapshot.
+"""
+from __future__ import annotations
+
+import argparse
+import copy
+import re
+import shlex
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Any, Optional
+
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+DEFAULT_WORKFLOW = REPO_ROOT / ".github" / "workflows" / "cache-budget-audit.yml"
+AUDIT_JOB_ID = "audit"
+CALIBRATION_STEP_NAME = "Calibrate the audit"
+LIVE_AUDIT_STEP_NAME = "Audit the live cache budget"
+LIVE_AUDIT_COMMAND = ["node", ".github/scripts/run-cache-budget-audit.mjs"]
+REQUIRED_PERMISSIONS = {"actions": "read", "contents": "read"}
+REQUIRED_TRIGGERS = ("schedule", "workflow_dispatch", "push")
+REQUIRED_WATCHED_PATHS = {
+    ".github/scripts/audit-cache-budget.mjs",
+    ".github/scripts/audit-cache-budget.test.mjs",
+    ".github/scripts/run-cache-budget-audit.mjs",
+    ".github/workflows/cache-budget-audit.yml",
+    ".github/workflows/ci.yml",
+    ".github/workflows/merge-readiness.yml",
+}
+
+
+class WorkflowLoader(yaml.SafeLoader):
+    """SafeLoader with YAML 1.2 boolean spelling for GitHub Actions keys."""
+
+
+# PyYAML's YAML 1.1 resolver turns the Actions key ``on`` into ``True``.  GitHub
+# Actions uses YAML 1.2 spelling, where only true/false are booleans, so retain
+# those values while treating ``on`` as the string key it is in the workflow.
+WorkflowLoader.yaml_implicit_resolvers = copy.deepcopy(yaml.SafeLoader.yaml_implicit_resolvers)
+for first_character, resolvers in WorkflowLoader.yaml_implicit_resolvers.items():
+    WorkflowLoader.yaml_implicit_resolvers[first_character] = [
+        resolver for resolver in resolvers if resolver[0] != "tag:yaml.org,2002:bool"
+    ]
+WorkflowLoader.add_implicit_resolver(
+    "tag:yaml.org,2002:bool",
+    re.compile(r"^(?:true|True|TRUE|false|False|FALSE)$"),
+    list("tTfF"),
+)
+
+
+def load_workflow(path: Path) -> Any:
+    """Parse a workflow with the GitHub Actions-compatible safe loader."""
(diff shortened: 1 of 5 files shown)


rizumita:
## Both review findings fixed, verified independently

`a205b72`. Measured by me, not taken from the implementer's report.

### Finding 1 — every rejection reason is now pinned

```
before   18 errors.append branches   16 mutation cases   17 tests
after    17 errors.append branches   —                   37 tests
```

The branch count dropped by one because the unreachable `"named live step must be a mapping"` branch was
removed rather than left uncovered. That matters: an unreachable branch and an unpinned branch look identical
to the next reader, so leaving it would have made the mutation table's completeness unverifiable by inspection.

Test count went 17 → 37. The per-item derivation from `validator.REQUIRED_PERMISSIONS`,
`REQUIRED_TRIGGERS` and `REQUIRED_WATCHED_PATHS` is retained, so widening what the validator protects still
widens the attack automatically; the structural and ordering mutations were added alongside it rather than
replacing it.

### Finding 2 — malformed input produces diagnostics, not tracebacks

All three shapes now report in the validator's own format:

```
--workflow /nonexistent.yml   cache-budget-audit workflow wiring: FAIL -- workflow file '/nonexistent.yml' does not exist
--workflow <invalid YAML>     cache-budget-audit workflow wiring: FAIL -- workflow YAML is invalid: while parsing a flow sequence
--workflow <YAML list>        cache-budget-audit workflow wiring: FAIL -- workflow must be a mapping
```

For the record on severity, since the original review phrasing could be read as stronger than it was: the
pre-fix `FileNotFoundError` traceback exited **nonzero**, so it never silently passed and was not a soundness
hole. It was a diagnostic-quality defect — someone whose workflow file was renamed got a Python stack trace
instead of being told the file was missing, which reads as "the validator is broken" rather than "the input is
wrong".

### Calibration re-verified

Replacing the missing-file diagnostic with an empty error list makes exactly
`test_cli_reports_missing_workflow_file` fail — **1 failed, 36 passed** — and restoring it returns **37/37**
with a clean tree. So the new diagnostics are individually pinned, not covered by an umbrella assertion.

Scope is unchanged: `.github/ruleset-contract.json` and `docs/DESIGN-ci.md` are still untouched, the new
context is still not required, and `src/fslc/` is unchanged.

PR #840, opened 2026-08-20
fix(ci): report cache budget audit failures through a privilege-separated reporter

### LINKED ISSUE #795: meta(ci): a failing cache budget audit files no issue, so #747's recurrence ran unobserved for four days

rizumita:
## Problem

The scheduled cache budget audit detects a real condition and then tells nobody:
`.github/workflows/cache-budget-audit.yml` ends immediately after invoking the runner, with no reporting step
and no downstream reporter.

Measured consequence: it failed on 2026-08-17 and again on 2026-08-18 at 9.27 GiB of the 10 GiB budget (93%,
above its 85% threshold). Nothing filed an issue, and #747's recurrence ran unobserved for **four days** until a
human happened to look. The audit was working correctly the whole time.

## Contract change: a privilege-separated reporter, not `issues: write` on the audit

The audit declares exactly `actions: read` and `contents: read`. Adding issue-write there would give a workflow
that *observes* the repository the power to *mutate* it — on a workflow that also accepts
`workflow_dispatch`. **The audit workflow is unchanged in this PR** and its permissions remain read-only;
verified.

Instead, `.github/workflows/cache-budget-audit-reporter.yml` subscribes to completed `cache budget audit` runs
via `workflow_run`, with `actions: read`, `contents: read`, `issues: write` and nothing else. It checks out the
**default branch, never the triggering SHA**, and acts only for this repository's default-branch `schedule`,
`push` or `workflow_dispatch` runs.

This follows `.github/workflows/post-merge-ci-reporter.yml`, the established pattern here, rather than
inventing a second one. Its dedup semantics come from the same source
(`.github/scripts/report-post-merge-ci.mjs:28`-`:49` stable marker, `:182`-`:215` deduplicate and reopen,
`:217`-`:247` close only on recovery; `docs/DESIGN-ci.md:1038`-`:1065` records the contract):

| Event | Behavior |
|---|---|
| first failure | files one canonical issue keyed by a stable hidden marker |
| same run reported again | idempotent — no second issue, no duplicate comment |
| later distinct failure | one occurrence comment on the existing issue |
| success | closes it |
| recurrence after close | reopens the same issue |
| stale out-of-order event | does not reopen health that has already recovered |

`cancel-in-progress: false` serializes reporters, and reconciliation runs against the latest completed trusted
audit before any mutation.

## Two layers of control, because one would recreate the defect

A reporter that silently stops reporting is exactly the failure this issue is about, one level up.

**Layer 1 — the reporter's own logic**, `.github/scripts/report-cache-budget-audit.test.mjs`, 47 Node tests
covering every row of the table above. **Wired into `tools/check-merge-readiness.sh`'s `check_automation()`
lane**, where `report-post-merge-ci.test.mjs` already lives — that lane is Node/stdlib-only, so this fits it
without disturbing its dependency contract. Writing the tests without wiring them is the exact shape this
repository has repeatedly mistaken for a gate (see the corrections recorded on #746, #786 and #802), so the
wiring is part of the change, not a follow-up.

**Layer 2 — the reporter's own wiring**, by extending #839's validator rather than adding a second one:

```
                                    before   after
errors.append branches in validator    17       61
tests collected in test_cache_bud...   37      137
```

It validates the `workflow_run` source and types, the exact permission set — including that `issues: write` is
present on the reporter and **absent from the audit** — the trusted/default-branch condition,
`cancel-in-progress: false`, safe checkout of the default branch rather than the triggering SHA, and an
unsuppressed runner command.

## Verified independently, not taken from the implementer's report

```
node --test report-cache-budget-audit.test.mjs        6 pass, 0 fail
pytest tests/test_cache_budget_audit_workflow.py     66 passed
cache-budget-audit.yml permissions                    actions: read, contents: read   (no issues: write)
```

Calibration probe: changing `REPORTER_PERMISSIONS` to drop `issues: write` from the *expected* set makes
**25 tests fail**; restoring it returns 66/66 with a clean tree. So the reporter's wiring assertions are
actually load-bearing rather than passing vacuously.

## Deliberately out of scope

No context is made required; `.github/ruleset-contract.json` and `docs/DESIGN-ci.md`'s required-context list
are untouched. `src/fslc/` is unchanged.

The audit's inability to distinguish reclaimable superseded duplicates from genuine growth remains #835 —
this PR reports the condition, it does not diagnose the cause.

`changelog.d/795-cache-budget-audit-reporter.fixed.md`.

Closes #795.

---

## Review history, and why these numbers moved

The counts above are re-measured against the current head, not the ones this
description was first written with. They grew because **every one of ten review
rounds found a real defect**, including one High (a `checkout.with.repository`
swap that would execute the head repository's script under the base
repository's `issues: write` token) and two that would have restored the very
failure mode this PR exists to fix — an audit rename silently disabling
reporting, and a same-name workflow impersonating the audit.

Rounds 9 and 10 are worth recording precisely, because round 9's blocking
finding was **rejected**:

- **Round 9** reported the non-required workflow-name uniqueness check as a
  blocker. The facts were right; the framing was not.
  `.github/workflows/cache-budget-audit-wiring.yml:5-19` already records, as an
  accepted decision, both that the workflow depends on PyYAML and so stays out
  of the stdlib-only automation-contracts lane, and that "This context is
  intentionally not required yet." Reading that header is what surfaced two
  real problems instead: it pointed at **closed** issue #802 as the rollout
  tracker (now #845, and the reference is updated here), and this PR does wire
  its reporter controls into the required lane, so the gap is much narrower
  than "none of these controls are required."
- **Round 10** found no actionable defect. It verified fail-closed behavior by
  execution, including the case that matters most: a malformed sibling **plus** a
  valid same-named impostor produces *both* the malformed-YAML error and the
  duplicate-name error, so an impostor cannot hide behind unparseable YAML.

### Calibration, checked by reverting the fix

The two new controls were confirmed to detect what they exist to detect.
Reverting only the validator hunk and re-running them gives 1 failed / 1
passed: the malformed-sibling control fails with empty stdout and a traceback on
stderr, while the well-formed-sibling accepting control still passes. This was
done twice independently.

### Residual risk, not verified by execution

`workflow_run` subscribes by workflow **display name**
(`workflows: ["cache budget audit"]`), and the uniqueness check compares that
name **exactly**. Whether GitHub's matching is case-sensitive cannot be
established without committing differently-cased default-branch workflows and
observing a live `workflow_run`, so it was not established here. Filed as #851,
which notes that the check can be made case-insensitive to fail closed
regardless of GitHub's behavior — no live experiment required.

Also filed from this work: #847 (a **scheduled** product-gate failure is always
skipped by the post-merge reporter, so an upstream drift that reddens `main` is
never filed — the adjacent reporter added here gets this right by including
`schedule` and `workflow_dispatch`, which makes the asymmetry visible).

### DIFF
--- a/.github/scripts/capture-cache-budget-audit-summary-fixture.mjs
+++ b/.github/scripts/capture-cache-budget-audit-summary-fixture.mjs
@@ -1,0 +1,112 @@
+#!/usr/bin/env node
+// SPDX-License-Identifier: Apache-2.0
+
+import { execFileSync } from "node:child_process";
+
+const [commit] = process.argv.slice(2);
+if (!/^[0-9a-f]{40}$/.test(commit ?? "")) {
+  throw new Error("usage: capture-cache-budget-audit-summary-fixture.mjs <40-hex-writer-commit>");
+}
+
+class FixtureClient {
+  constructor(completedRuns) {
+    this.completedRuns = completedRuns;
+    this.issues = [];
+    this.comments = [];
+    this.labels = new Set();
+    this.nextIssue = 100;
+    this.nextComment = 1000;
+  }
+
+  async listCompletedWorkflowRuns(_workflowId, _branch, page) {
+    return this.completedRuns.slice((page - 1) * 100, page * 100);
+  }
+
+  async ensureLabel(label) {
+    this.labels.add(label);
+  }
+
+  async listIssues(_label, page) {
+    return this.issues.slice((page - 1) * 100, page * 100);
+  }
+
+  async listIssueComments(number, page) {
+    return this.comments
+      .filter((comment) => comment.issue === number)
+      .slice((page - 1) * 100, page * 100);
+  }
+
+  async createIssue(issue) {
+    const created = { ...issue, number: this.nextIssue, state: "open" };
+    this.nextIssue += 1;
+    this.issues.push(created);
+    return created;
+  }
+
+  async createIssueComment(number, body) {
+    this.comments.push({
+      id: this.nextComment,
+      issue: number,
+      body,
+      user: { login: "github-actions[bot]" },
+    });
+    this.nextComment += 1;
+  }
+
+  async updateIssueComment(commentId, body) {
+    this.comments.find((comment) => comment.id === commentId).body = body;
+  }
+
+  async deleteIssueComment(commentId) {
+    this.comments.splice(
+      this.comments.findIndex((comment) => comment.id === commentId),
+      1,
+    );
+  }
+
+  async updateIssue(number, update) {
+    Object.assign(this.issues.find((issue) => issue.number === number), update);
+  }
+}
+
+function workflowRun(overrides = {}) {
+  return {
+    id: 41,
+    workflow_id: 7,
+    run_number: 12,
+    run_attempt: 1,
+    conclusion: "failure",
+    event: "schedule",
+    head_branch: "main",
+    head_repository: { full_name: "ymm-oss/fsl" },
+    head_sha: "0123456789abcdef0123456789abcdef01234567",
+    html_url: "https://github.com/ymm-oss/fsl/actions/runs/41",
+    ...overrides,
+  };
+}
+
+// This is deliberately a capture tool, not a test dependency. Regeneration
+// requires the original writer commit to be locally available; that deliberate
+// prerequisite prevents a squash merge from silently changing test inputs.
+const source = execFileSync(
+  "git",
+  ["show", `${commit}:.github/scripts/report-cache-budget-audit.mjs`],
+  { encoding: "utf8" },
+);
+const { reconcileCacheBudgetAudit } = await import(
+  `data:text/javascript;base64,${Buffer.from(source).toString("base64")}`,
+);
+const client = new FixtureClient([workflowRun({ id: 41, run_number: 12 })]);
+await reconcileCacheBudgetAudit({
+  client,
+  repository: "ymm-oss/fsl",
+  defaultBranch: "main",
+  workflowRun: workflowRun({ id: 42, run_number: 13 }),
+});
+const summary = client.comments.find((comment) =>
+  comment.body.includes("cache-budget-audit-occurrence-summary"),
+);
+if (!summary) {
+  throw new Error("historical writer did not produce an occurrence summary");
+}
+process.stdout.write(`${summary.body}\n`);
--- a/.github/scripts/capture-cache-budget-audit-summary-fixture.mjs
+++ b/.github/scripts/capture-cache-budget-audit-summary-fixture.mjs
@@ -109,4 +109,4 @@ const summary = client.comments.find((comment) =>
 if (!summary) {
   throw new Error("historical writer did not produce an occurrence summary");
 }
-process.stdout.write(`${summary.body}\n`);
+process.stdout.write(summary.body);
--- a/.github/scripts/fixtures/cache-budget-audit-summary-fixtures.mjs
+++ b/.github/scripts/fixtures/cache-budget-audit-summary-fixtures.mjs
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// These are actual outputs captured from historical writers, not reconstructed
+// test text. Regeneration deliberately requires the source commit named below:
+// use the recorded capture command only after making that original object
+// available, then update its writer/output digests in the same review.
+export const HISTORICAL_SUMMARY_FIXTURES = Object.freeze([
+  Object.freeze({
+    id: "original-unqualified",
+    provenance: Object.freeze({
+      writerCommit: "cbb00dca5acf99742743a22dd33affa29378d85e",
+      writerSha256: "3dc17ad18cd035bb9ef197742e283d47e4d6d00169941255aef837cb673185e9",
+      captureCommand: "node .github/scripts/capture-cache-budget-audit-summary-fixture.mjs cbb00dca5acf99742743a22dd33affa29378d85e",
+      outputSha256: "6ea77d6d6b95de9f6acbfec0a00cfd4d13aae0135d436969aa1de34d2937d649",
+    }),
+    body: `<!-- cache-budget-audit-occurrence-summary -->
+<!-- cache-budget-audit-occurrence:42:1 -->
+<!-- cache-budget-audit-cursor:13:1 -->
+Detailed recurrence comments are capped at 20; this rolling summary records the latest recurrence.
+Coalesced failed attempts: 1.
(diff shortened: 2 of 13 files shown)


rizumita:
## Review round 1: five findings, all fixed. `82465a1`.

### Finding 1 (Medium) — the layer-2 validator did not enforce the security boundary this PR claims

I reproduced it before the fix, driving `validate_reporter_workflow` through the tests' own loader:

```
baseline                                                  []  correctly accepted
job-level permissions -> actions/contents/issues: write    []  ACCEPTED
arbitrary shell step inserted before the runner            []  ACCEPTED
checkout changed to actions/checkout@main                  []  ACCEPTED
```

Only top-level permissions were checked. That is not the effective permission set — GitHub applies
job-level permissions **after** workflow-level ones. And a step inserted before the runner can modify the
checked-out reporter script, which the next step then executes **with the write token**. The checked-in
workflow was safe; the control was hollow, which is exactly the defect class this PR exists to prevent,
reproduced inside the mechanism meant to prevent it.

Re-measured after the fix, plus one variant I added myself:

```
job-level permissions        rejected: "reporter job must not declare a permissions override"
step inserted before runner  rejected: "reporter job steps must be exactly checkout, setup-node, ..."
checkout @main               rejected: "reporter checkout action must be pinned to the approved commit"
step appended at the end     rejected: "reporter job steps must be exactly ..."
```

### Finding 2 (Medium) — idempotence broke past one comment page

`report-cache-budget-audit.mjs` fetched only `per_page=100, page 1`. With the occurrence marker as comment
101, reconciling the identical run/attempt produced comment 102, so "the same run is a no-op" held only until
an issue reached 100 comments. The same assumption applied to issues and workflow runs.

Fixed on both axes rather than just the first: all three API reads now go through a `listAllPages` helper
(`:135`, `:165`, `:179`), **and** the unbounded occurrence log is gone — the reporter now maintains a single
updated summary comment (`occurrenceSummaryMarker` / `recoverySummaryMarker`) instead of appending one per
failure. That removes the growth problem rather than paginating around it.

### Finding 3 (Medium) — the contract had no accepted design record, and my brief caused that

`docs/DESIGN-ci.md` gains a "Cache-budget audit failure reporting" section recording why the audit stays an
observer, the reporter's exact least-privilege set, its trust filter (same repository, default branch, one of
push/schedule/workflow_dispatch), default-branch checkout rather than the triggering SHA, the pinned-and-exact
step sequence and *why* — "so a mutable action or an inserted pre-run shell step cannot alter the checked-out
reporter before it receives the write token" — the canonical marker, and the occurrence/recovery policy.

**This gap was my fault.** My implementation brief said not to touch `docs/DESIGN-ci.md`. I meant its
required-context list, which this PR still must not change, and I stated it too broadly. Documenting the
reporter's contract is a separate obligation that `AGENTS.md` requires, and my instruction had excluded it.

### Findings 4 and 5 (Low)

The audit's watched paths now include the reporter workflow, its script and its test, so a push touching them
exercises the deployed path instead of waiting for the next schedule. And marker matching now filters on the
bot author, so a commenter who races the predictable `run_id:run_attempt` marker can no longer suppress a
reporter-authored comment — that corrupted the audit trail without affecting the state machine.

### Counts, measured

```
                                      before   after round 1
validator errors.append branches         35         39
tests/test_cache_budget_audit_...        66         75
Node reconciliation tests                 6         10
```

`.github/ruleset-contract.json` is still untouched and no context is made required.

What round 1 confirmed as already correct: no fork-to-`issues: write` exploit in the checked-in workflow; the
trust filter; ordering by `(run_number, run_attempt)` rather than timestamps; `run_id:run_attempt` as the
re-run identity (a re-run is correctly a distinct occurrence); marker-as-identity; human-close-then-reopen; and
that the merge-readiness wiring genuinely propagates failure — an injected reporter-test failure exited the
lane with status 17.

Round 2 review is running.

PR #854, opened 2026-08-20
chore(ci): pin the Rust toolchain and control the pin

rizumita:
## Problem

`main` went red on 2026-08-20 **with no repository change**.

`rustc`/`clippy` 1.98.0 was released 2026-08-18. All eleven audited workflow
references used the floating `dtolnay/rust-toolchain@stable`, so the scheduled
`product gate` picked up two new lints that this repository's `-D warnings`
turns into errors, on a zero-line diff:

| `product gate` run | time | event | result |
|---|---|---|---|
| `32350147121` | 08-20T08:42 | push | success |
| `32406084360` | 08-20T18:58 | **schedule** | **failure** |

#846 fixed the lints. #847 fixed the reporting half — nothing had filed an
issue. **This PR removes the cause**, so the next upstream release does not
repeat it.

## Change

Pin all eleven audited references to `@1.98.0`. `stable` resolves to 1.98.0
today — verified against
`https://static.rust-lang.org/dist/channel-rust-stable.toml`, whose
`[pkg.rustc] version = "1.98.0 (88d9e12ae 2026-08-18)"` matches the commit hash
of the toolchain this was tested on — so the compiler that actually runs does
not change.

**Correction:** an earlier version of this description also claimed
`Swatinem/rust-cache` keys are unaffected. **That is wrong**, measured on this
PR against #853, which leaves `@stable` in place:

| PR | ref | cache key |
|---|---|---|
| #853 | `@stable` | `v0-rust-rust-workspace-Linux-x64-`**`0b9fd15e`**`-9efe1eb7` |
| #854 | `@1.98.0` | `v0-rust-rust-workspace-Linux-x64-`**`718c915e`**`-9efe1eb7` |

The lockfile component is identical; the environment component differs, because
the "Rust Versions" list `rust-cache` hashes differs. `@stable` lists only
1.98.0; `@1.98.0` lists the runner image's preinstalled 1.97.1 **as well as**
1.98.0. See the comment thread for the raw output.

A cache miss is slower, never wrong, and it is transient: once this merges,
`main`'s own runs produce the same two-version list and the saved caches realign.
`merge-readiness.yml`'s restore-only match against `ci.yml` still holds, since
both are pinned to the same ref — which is exactly what its `:43`/`:99` comments
claim. The residual cost is that the cache key now depends on the image's
preinstalled Rust, which is itself unpinned, so an image bump will invalidate
caches where it previously would not.

| file | references |
|---|---|
| `ci.yml` | 8 |
[...]

PR #853, opened 2026-08-20
fix(ci): report scheduled and dispatched product gate failures, not only pushes

rizumita:
## Problem

`main` was red for hours and **nothing filed an issue**. It was found by manual
inspection while investigating something else.

The scheduled `product gate` failed and the reporter for the *same commit* was
skipped:

| run | workflow | event | `head_sha` | result |
|---|---|---|---|---|
| `32406084360` | product gate | **schedule** | `abef68c` | **failure** |
| `32406638065` | post-merge CI reporter | workflow_run | `abef68c` | **skipped** |

The last eight reporter runs were all `skipped`. The cause was
`post-merge-ci-reporter.yml` requiring `workflow_run.event == 'push'`.

The daily cron is the **only** mechanism that revalidates `main` without a new
commit — so it is the only detector of upstream drift: a new compiler, a new
transitive dependency, an expired cache. Its entire output was being dropped.
The live instance: `rustc`/`clippy` 1.98.0 released 2026-08-18, `main` went
green-to-red with a zero-line diff, and nothing reported it (fixed by #846).

## Contract conflict, resolved

`docs/DESIGN-ci.md` contradicted itself, and resolving that is most of this PR:

- **:1123** documented the reporter as acting "only when the triggering run was
  a push to the repository's default branch."
- **:172** claimed a wrongly exempted defect "surfaces through the existing
  post-merge reporting contract below" — for the `main` push, **schedule** and
  dispatch triggers.

Resolved in the *widen-the-reporter* direction, and both passages now say the
same thing. There was already a precedent in the same document: **:896**
describes the cache-budget-audit reporter (merged in #840) as requiring "the same
repository, the default branch, and one of `push`, `schedule`, or
`workflow_dispatch`". This aligns the older reporter with the newer one rather
than inventing a second shape.

## The part that is not just the workflow `if`

The push assumption lived in **three** places, and widening only one would have
reproduced the "nothing gets filed" symptom by a different route:

| # | location | before |
|---|---|---|
| 1 | `post-merge-ci-reporter.yml` job `if` | `event == 'push'` |
| 2 | `report-post-merge-ci.mjs` `main()` guard | `event !== "push"` → return |
| 3 | `report-post-merge-ci.mjs` `latestCompletedWorkflowRun` | `&event=push` **hardcoded in the API query** |

(3) was the trap. `reconcilePostMerge` replaces the triggering run with that
"latest completed trusted run" when `isNewerRun` says so. Widen only the `if`,
[...]

PR #863, opened 2026-08-21
fix(ci): parse workflow YAML instead of scanning lines

### LINKED ISSUE #856: toolchain 固定監査は YAML パースではなく行スキャンのため、妥当な YAML の書き方で素通りする穴が出続ける(4件実測)
# Problem

`.github/scripts/audit-toolchain-pin.mjs`（PR #854）は **YAML をパースせず行スキャンで**
`uses:` 参照を集める。そのため「妥当な YAML だが正規表現が見ない」書き方が
そのまま監査を素通りする。PR #854 のレビュー中に**4件**見つかり、
すべてプローブ（変異を当てる）でのみ検出できた。

# 実測（PR #854 レビュー中、2026-08-21）

最初の実装は次で固定していた:

```js
^\s*-\s*uses:\s*dtolnay\/rust-toolchain@(\S+)\s*$
```

| # | 書き方 | 素通りした理由 | 現在の分類 |
|---|---|---|---|
| 1 | `- uses: "action@stable"` | 引用符を剥がしていなかった | `floating-channel` |
| 2 | `- uses: action@${{ matrix.tc }}` | `(\S+)\s*$` が式の空白で不一致 | `expression-ref` |
| 3 | `- name: x` + 兄弟 `  uses: action@stable` | リストのハイフンを必須にしていた | `floating-channel` |
| 4 | `- uses : action@stable` | コロン前の空白を許していなかった | `floating-channel` |

4 は YAML として妥当であることをパーサで確認した:

```
>>> yaml.safe_load("steps:\n  - uses : action@stable")
step keys: ['uses']   uses value: 'action@stable'
```

**4件すべてで、コミット済みワークフローへの判定は変わらなかった。**
それが「穴」である証拠である（実態と食い違っていたのではなく、見えていなかった）。
プローブディレクトリでの audited 件数は 12 → 15 に増えた。

# なぜ根絶したいか

穴は塞いだが、**行スキャンである限り5つ目が無いと言える根拠が無い。**
思いつく未検証の候補:

- 複数行フロースカラー、アンカー/エイリアス（`*ref`）
- reusable workflow 呼び出し（`uses:` が workflow を指し、その中で toolchain を入れる）
- composite action 経由の間接的なインストール
- `rust-toolchain.toml` / `RUSTUP_TOOLCHAIN` による実行時上書き
  （これは `uses:` 参照の話ではないので、そもそも別の検出面）

`audit-toolchain-pin.mjs` の `USES_KEY_PATTERN` のコメントには、
「これは行スキャンでありパースではない、だから各々が穴だった。
5つ目の緩和が必要になったらパースに移行せよ」と記録してある。本 issue はその移行。

# 制約: どこで走らせるか

`tools/check-merge-readiness.sh` の `check_automation()` は
**Node/stdlib のみ**という依存契約を持つ（同ファイルの `python3 -c` の但し書き、
および `.github/workflows/cache-budget-audit-wiring.yml:5-14` が
PyYAML を third-party として必須 lane の外に置いた理由を参照）。

したがって選択が必要である:

**案 A: 必須 lane に YAML パーサ依存を入れる。**
Node の場合 `yaml` パッケージ等。stdlib-only 契約を変更する決定になるので、
`docs/DESIGN-ci.md` に根拠を記録すること。

**案 B: 非必須ワークフローに出す。**
`cache-budget-audit-wiring.yml` と同じ形。ただしそれは #845 が扱っている
「必須でない制御」の問題をもう一つ作ることになる。**推奨しない。**

**案 C: パースせずに堅牢化する。**
GitHub Actions の `uses:` に現れうる構文を列挙し、
それぞれに rejecting control を与える。根絶ではなく被覆の拡大。

# 期待される挙動

いずれの案でも:

1. 上の4件が引き続き拒否されること（既存の18対照を退行させない）。
2. 上の「未検証の候補」の各々について、拒否するか、
   **明示的に対象外と宣言して理由を書く**こと。
   黙って見ないのが今回の欠陥だった。
3. `audited` 件数の下限（現在 `MINIMUM_AUDITED_REFERENCES = 11`）を維持し、
   監査対象ゼロで PASS しないこと。これもレビューで見つかった欠陥である。

# 参考

- PR #854（固定と行スキャン監査、コミット `cd4cc7f` と `cddc70f` が穴の修正）
- #845（必須でない制御を必須化するロールアウト）
- `.github/workflows/cache-budget-audit-wiring.yml:5-14`（stdlib-only lane の根拠）

# 影響範囲

`.github/scripts/audit-toolchain-pin.mjs`,
`.github/scripts/audit-toolchain-pin.test.mjs`,
`tools/check-merge-readiness.sh`, `docs/DESIGN-ci.md`

rizumita:
## Problem

Two CI controls read workflow YAML with **line-based regular expressions**. Between
them they were rewritten **eighteen times** — fourteen for the toolchain audit,
four for the reporter condition — by three people. Every hole was a place where a
regex assumption diverged from YAML's grammar: quoted values, comments, folded
lines, indentation width, flow mappings, blank lines inside a block scalar,
`uses :` with a space before the colon, `env:` accepted as `with:`.

**In all eighteen the control's verdict on the committed workflows stayed
unchanged and CI stayed fully green.** None was visible without probing.

Two findings settle the argument. Given

```yaml
with: {
  components: rustfmt, # , toolchain: 1.88.0 }
  toolchain: 1.89.0
}
```

the scanner read `1.88.0` **out of a comment**; PyYAML says `1.89.0`. Given
`with: {toolchain: 1.88.0, components: "rustfmt}"}` its brace counting found no
input at all; PyYAML says `1.88.0`. A parser answers both without being told
about either.

### The mistake underneath

**This repository already parsed workflow YAML with PyYAML**, in
`.github/scripts/validate-cache-budget-audit-workflow.py`, and `pyproject.toml`
declares `PyYAML>=6`. The `Node/stdlib-only` comment in
`tools/check-merge-readiness.sh` was read as "no parser available" when it meant
"no third-party dependency in the fast lane" — so a hand-written parser was
patched eighteen times **in order to avoid a dependency**. That trade is
backwards, and it is the substance of this issue.

## Change

Two PyYAML validators replace the scanners, and the required lane gains PyYAML.

### `validate_toolchain_pin.py`

Pins all eleven development references to `@1.98.0` — this also supersedes #854,
now closed — and holds `release.yml` to an MSRV **contract** rather than
exempting it: the action must be pinned by 40-character commit SHA and its
`with:` `toolchain:` must name `rust/Cargo.toml`'s `rust-version`, read at run
time.

It walks the composed node tree, so a step is found wherever it legitimately sits
and however it is spelled. `yaml.compose` rather than `yaml.load`, because node
source marks keep `FILE:LINE` in the diagnostics — the scanner partly existed to
preserve those, and parsing does not have to cost them.

Beyond `jobs.steps`, it also audits **local reusable workflows and local
composite actions**. External reusable workflows are outside repository source
authority and are stated as such rather than silently skipped.

### `validate_post_merge_reporter_workflow.py`

The reporter holds `issues: write`, so it gets a deliberately narrow, statically
verifiable shape. The implementer's own recommendation, adopted after being asked
for reasoning before code: **reject run-time indirection, and make
`jobs.reconcile` a contract boundary**. Rejected alternatives, with their reasons
recorded on the issue: fixing the job ID alone does not prove the writer was not
moved elsewhere, and a run-time check cannot block a bad change pre-merge.

### The single source of truth

`.github/scripts/trusted-workflow-events.json` is read by **both**
`report-post-merge-ci.mjs` at run time and the Python validator. The YAML gate
could not import a JavaScript constant, and that gap is what every previous
revision of the reporter control fell through. Now neither side can change alone.

## Test evidence

| gate | result |
|---|---|
| `validate_toolchain_pin.py` | exit 0, 11 references |
| `pytest tests/test_toolchain_pin.py` | exit 0, **22 passed** |
| `pytest tests/test_post_merge_reporter_workflow.py` | exit 0, **12 passed** |
| `node --test report-post-merge-ci.test.mjs` | exit 0, 12 passed (runtime behaviour retained) |
| `./tools/check-merge-readiness.sh automation` | exit 0 |
| `actionlint` on all four changed workflows | exit 0 |
| `./tools/aggregate_changelog.sh check` | exit 0 |

### Every bypass runs as an actually-failing fixture

Not asserted — executed. For the reporter contract: comment counterfeit,
`&& false`, blank-line suffix, renamed-writer decoy, `env:`/shell indirection,
local composite invocation, and a second writer job. For the toolchain pin:
floating and expression refs, composite and reusable paths, `env:` and nested
inputs, duplicate keys, invalid YAML, a missing MSRV step, and the vacuous-audit
floor.

## Not lost, and deliberately checked

This branch was first cut from `main` **before #853 merged**, so its diff briefly
contained a 494-line reversion of that fix — which would have restored the outage
where scheduled product-gate failures went unreported. After rebasing, verified:
`changelog.d/847-scheduled-gate-reporting.fixed.md` still exists, and
`post-merge-ci-reporter.yml` retains the `schedule` predicate and the
`head_repository` fork guard #853 added.

## Required lane now depends on PyYAML

`merge-readiness.yml`'s automation job gains `actions/setup-python` and
`pip install ".[dev]"`, and `docs/DESIGN-ci.md` records why: the audit has to be
**both correct and required**. Putting it in a non-required workflow — the
alternative that would have avoided the dependency — would have made it weaker
than the line scanner it replaces.

Closes #856


rizumita:
## Reviewer High resolved — independently mutation-probed

The review's High was that `validate_post_merge_reporter_workflow.py` validated `run:` strings and
local composites but nothing about the privileged job's checkout step. Grepping the original file for
`checkout|repository|persist-credentials|ref` returned only three hits, all inside the `if:` condition
string. A `repository: other/repo` addition would have let a job holding `issues: write` execute
another repository's code.

`d887afe` brings the validator to the approved-mapping standard already used by
`.github/scripts/validate-cache-budget-audit-workflow.py`: the checkout action is pinned by SHA, `ref`
is pinned to the default-branch expression, `persist-credentials: false` is required, and
`_reject_unapproved_keys` is now called at 22 sites. That last point is the substance — an unapproved
key is rejected because it is not on the approved list, not because someone predicted it.

I re-probed each fail-open path myself against the working tree rather than relying on the added
tests. Baseline `exit 0`; each single mutation applied in isolation and reverted:

| mutation | exit | first diagnostic |
|---|---|---|
| add `repository: octocat/Hello-World` to checkout | 1 | `checkout configuration must not declare unapproved keys: repository` |
| `persist-credentials: false` → `true` | 1 | rejected |
| checkout `ref` → `github.event.workflow_run.head_sha` | 1 | rejected |
| `actions/checkout@34e1148…` → `@v4` | 1 | rejected |
| drop the `head_repository.full_name == github.repository` guard | 1 | rejected |
| add a job-level `permissions:` override | 1 | rejected |

Baseline returned to `exit 0` after restoring the file, so these are detectors and not a validator
that fails on everything.

The `ref` row matters most: pinning to `head_sha` is the untrusted-code path this workflow exists to
avoid, and it is now rejected rather than merely conventional.

Issue #472, opened 2026-07-24
fix(testgen): 本物の violated 反例を exit-2 spec error に変換してトレースを失う問題を直す


rizumita:
# Problem

契約違反（exit code + JSON envelope + docs の約束）。A real `violated` / `reachable_failed` verdict is misclassified as an exit-2 spec `error` with `kind:"semantics"` and the replayable counterexample evidence is destroyed. Three consequences: (1) exit-code contract break — callers gating on "1 = spec bug, 2 = fix your input" misroute every violated spec; (2) evidence loss — the trace/blame/`unreached[].classification` that the repair loop and agent skills depend on is replaced by an adapter-internal message with no `loc` and no spec name; (3) `--strict` no longer does what LANGUAGE/DESIGN-bridge/CHANGELOG/skill reference all promise. Secondary: 検出力ゼロ — the only parity harness for testgen (`tools/check_rust_phase3_commands.py`) uses green specs only and is invoked by no CI gate, and status 3 (internal error, e.g. solver init failure inside run_scenarios_mode) is also silently downgraded to 2. Not false-proved: exit 2 is non-zero, so no confidently-green false negative — but it is a false attribution of a verification failure to the user's input.

# Evidence

Root cause (code): `rust/fslc/src/main.rs:9469-9472`
```
let (scenarios, status) = run_scenarios_mode(path, depth, deadlock_mode, !strict);
if status == 2 { return (scenarios, status); }
```
`run_scenarios_mode` (main.rs:4813-4825) delegates to `run_verify` when `result.violation.is_some() || result.leadsto_violation.is_some() || (!allow_unreached && unreached)`, i.e. it returns **status 1** with a `violated` / `reachable_failed` envelope. Only `status == 2` short-circuits, so the verify envelope is passed on as if it were a scenarios envelope; `fsl_tools::validate_scenarios` (rust/fsl-tools/src/testgen.rs:205-209) then fails on the missing `scenarios` array and `run_testgen` maps that to `semantic_error_output` (exit 2). `run_domain_testgen` inherits it (main.rs:6479 `if generic_status == 2`). Status 3 (internal) is likewise downgraded to 2.

Committed-corpus reproduction (`specs/cart_buggy.fsl`, invariant NoNegativeStock):
- `rust/target/debug/fslc verify specs/cart_buggy.fsl --depth 6` → exit 1, `result:"violated"`, `violation_kind:"invariant"`.
[...]

PR #866, opened 2026-08-22
fix(testgen): fail closed on a walk violation instead of baking a no-op expectation

### LINKED ISSUE #843: `testgen` の baked walk が Monitor の violation を捨て、契約に無い no-op 期待値を焼き込む（pytest と baked 5 target が不一致）
## Problem

`testgen_trace_vectors` は `Monitor::step` の `StepResult` を破棄し、直後の `monitor.state` を
`expected` として焼き込む。Monitor は違反時に rollback するため、**違反したステップは
「状態が変わらなかった」という期待値として記録される**。

これは言語契約のどこにも存在しない **no-op 意味論**である。結果：

- **適合実装はこのテストで落ちる**（偽陽性）
- **その地点で黙って何もしない実装は通る**（invariant 違反が表面化しない）
- **`--target pytest` は live Monitor を使うので同じ spec で `pytest.fail` する**
  → 同一入力に対し pytest と baked 5 target（vitest / swift / kotlin / dart / phpunit）が
  異なる結論を出す

`AGENTS.md` の「記号検証・具体 Monitor・solver 非依存 BFS の三者が一致しなければならない。
**自信を持って緑になる false negative はクラッシュより危険**」に直接触れる。

## Mechanism

`rust/fslc/src/lib.rs:340`：

```rust
monitor.step(choice).map_err(|error| error.to_string())?;
steps.push(json!({
    "action": display_name(&action),
    "params": params,
    "expected": ordered_object(&state_json(&monitor.state), &state_order)
}));
```

`map_err(...)?` は `Result` の Err だけを伝播する。`Ok(StepResult { violation: Some(_), .. })` は
そのまま捨てられる（`StepResult` は `rust/fsl-runtime/src/lib.rs:751-757` で
`violation: Option<Violation>` を持つ）。

walk の長さは同関数内で **`for _ in 0..100` と固定**されており、`--depth` と独立である。

## Why the existing fail-closed check does not cover it

`testgen` は違反 spec に対しては正しく fail-closed になる。違反が既定 depth 内にある入力では：

```console
$ fslc testgen violating.fsl --target vitest -o out.test.ts
{"result":"violated","violation_kind":"invariant","invariant":"AtMostTwo","violated_at_step":3, ...}
$ echo $?
1
```

生成物は作られない。**しかし検証は `--depth`（既定 8、`rust/fslc/src/main.rs:3313`）で打ち切られ、
walk は 100 ステップ固定である。** 違反がその間にある spec で穴が開く。

## Reproducer

```fsl
// 違反は step 30 でしか到達しない。testgen の既定検証深度より先にある。
spec LateViolation {
  type N = 0..30
  state { c: N }
  init { c = 0 }
  action inc() {
    requires c < 30
    c = c + 1
  }
  invariant AtMost29 { c <= 29 }
}
```

`fslc 4.3.0`（native-z3 / Z3 4.16.0.0、公式 Release macOS arm64、SHA-256 照合済み）での実測：

| コマンド | 結果 |
|---|---|
| `verify --depth 8` | `verified` |
| `verify --depth 30` | `violated` / `invariant AtMost29` / step 30 |
| **`testgen --target vitest`（既定 depth）** | **`generated` / exit 0** |
| `testgen --depth 30 --target vitest` | `violated`（正しく拒否） |

生成された `RANDOM_WALK` は 100 ステップで、`expected` の `c` は

```text
1, 2, 3, ..., 28, 29, 29, 29, ... 29   ← 約71ステップにわたり 29 のまま
```

つまり `{"action":"inc","params":{},"expected":{"c":29}}` が繰り返し焼き込まれる。
spec が言っているのは「`c = 29` で `inc()` は enabled（`requires c < 30`）であり、
その後継状態 `c = 30` は `AtMost29` に違反する」であって、
「`inc()` しても 29 のまま」ではない。

### target 間の不一致

同じ spec に対する `--target pytest`（`rust/fsl-tools` の pytest emitter）は walk を実行時に回す：

```python
result = mon.step(action, params)
if not result.get('ok'):
    pytest.fail(
        f'spec oracle violation at {action} {params}: '
        f"{result.get('kind')} {result.get('name', '')}"
    )
```

したがって pytest は step 30 で `spec oracle violation at inc {}: invariant AtMost29` として落ちる。
baked target は落ちず、代わりに適合実装を落とす。**同一入力・同一 seed で結論が反対になる。**

## Impact

1. **偽陽性**：仕様どおり実装したコードが、baked target のテストで落ちる。
   利用者は実装を疑い、実際には spec 側に違反がある
2. **偽陰性（より危険）**：その地点で例外を投げず状態も変えない実装は緑になる。
   本来 surface すべき invariant 違反が、テストを通過した状態で隠れる
3. **target parity の破れ**：`docs/DESIGN-bridge.md` は6 emitter が同じ scenario 集合を
   別ハーネスへ描画する設計だが、この経路では pytest と他5つが同じ入力に対して一致しない
4. **証拠契約の汚染**：`testgen-trace.v1` は versioned contract として配布されるため、
   これを消費する外部ツールも同じ誤った期待値を受け取る

## Suggested direction

最小で健全な選択肢は次のいずれか。どれを採るかは maintainer の判断に委ねる。

**A. walk を violation で fail-closed に打ち切る（推奨）**

`testgen_trace_vectors` が `StepResult.violation` を検査し、`Some(_)` なら
trace 生成を **error** にする（`verify` が違反 spec で exit 1 になるのと同じ扱い）。
`kind` / `name` / そのステップ番号 / action と params を診断に含める。
「verified な spec からしかハーネスを作らない」という文書化済みの推奨
（`docs/LANGUAGE.md:1652-1655`）と整合する。

**B. walk 長を検証深度に揃える**

100 固定をやめ、`--depth` で検証済みの範囲を超えないようにする。
ただし walk が短くなり検出力が落ちる。A と併用が望ましい。

**C. violation を trace の一級市民にする**

`testgen-trace.v1` に `outcome` を追加し、baked emitter が
「このステップは違反する」ことを assert できるようにする。
schema major の変更と6 emitter の同時対応が必要で、コストは最大。
`fslc conformance` が既に outcome 付き vector を出しているので、
そちらとの役割分担を先に決める必要がある。

いずれの場合も、**`monitor.step` の戻り値を捨てる現在の書き方は残せない**。

## Required controls

- **拒否対照（必須）**：上記 `LateViolation` を入力に、選ばれた挙動を pin する。
  A を採るなら `testgen --target vitest` が **exit 非 0** になり、
  `c=29 → expected c=29` を含む生成物が作られないこと
- **受入対照**：違反が到達しない spec（例：`invariant AtMost30 { c <= 30 }`）では
  100 ステップの walk が従来どおり生成され、既存 golden が byte 一致すること
- **target parity 対照**：同一 spec・同一 seed で pytest と baked 5 target が
  同じ結論（生成拒否 / 生成成功）になること。現在は反対の結論になる
- `testgen-trace.v1` を消費する経路（6 emitter と外部消費者）への回帰

## Related

- `docs/DESIGN-bridge.md` — 6 emitter が同一 scenario 集合を共有する設計
- `docs/LANGUAGE.md:1652-1668` — 推奨ワークフロー（verify → testgen）と
  baked walk の設計意図
- `docs/DESIGN-kernel-contract.md:75-102` — Monitor の失敗分類順序と
  `attempted_state` の契約（violation を落とさない前提）
- `fslc conformance` は outcome 付きの bounded exhaustive vector を既に出しており、
  選択肢 C を検討する場合は役割の重複を先に整理する必要がある

## Notes

調査は read-only で行っており作業ツリーへの変更はない
（`git status --short` は空、HEAD `abef68c`）。再現用 `.fsl` は本文に全文を含めた。

### LINKED ISSUE #858: 検証結果を運ぶ公開型 33 件すべてに `#[must_use]` が無く、`Result` の must_use では穴が閉じない（#843 の実際の原因）
## Problem

`violation` / `outcome` を運ぶ公開型が **33 件あり、そのうち `#[must_use]` が付いているものは 0 件**である。
一方このコードベースには `#[must_use]` が **316 箇所**あるが、**すべて関数**に付いている
（314 件が `fn`、残り 2 件は doc 行の誤検出）。clippy `pedantic` の `must_use_candidate` が
関数を対象にするため、現状の規律は **lint 駆動であって「結果型を守る」設計になっていない。**

これは理論上の指摘ではない。**#843 はこの欠落の直接の帰結である。**

## The gap `Result` does not cover

`Result` は標準で `#[must_use]` なので、次のコードは **Result 側の義務を満たしてしまう**。

```rust
// rust/fslc/src/lib.rs:340
monitor.step(choice).map_err(|error| error.to_string())?;
```

`?` が Err を伝播した後に残るのは **内側の `StepResult`** であり、これが文の値として捨てられる。
`StepResult`（`rust/fsl-runtime/src/lib.rs:751-757`）は `violation: Option<Violation>` を持つため、
**違反が黙って捨てられる。**

宣言側にも `#[must_use]` は無い（`rust/fsl-runtime/src/lib.rs:951`）：

```rust
pub fn step(&mut self, enabled: &EnabledAction) -> Result<StepResult, RuntimeError>
```

仮に関数側に付けても `Result` に対して冗長であり、**この穴は閉じない。型に付ける必要がある。**

## Measured: the fix would have blocked #843 at compile time

最小再現をコンパイルして確認した。

```rust
#[derive(Debug)]
#[must_use]
pub struct StepResult { pub violation: Option<String> }

fn step() -> Result<StepResult, String> { Ok(StepResult { violation: None }) }

fn run() -> Result<(), String> {
    step().map_err(|e| e.to_string())?;   // #843 と同一の形
    Ok(())
}
```

```console
$ cargo build
warning: unused `StepResult` that must be used
 --> src/main.rs:9:5
  = note: `#[warn(unused_must_use)]` (part of `#[warn(unused)]`) on by default
```

CI の Rust ゲートは

```bash
cargo clippy --manifest-path rust/Cargo.toml --workspace --all-targets --locked -- -D warnings
```

なので、この警告は **マージ不可能なビルド失敗**になる。つまり #843 は**構造的に混入不可能だった。**

## Scope: 33 types, 0 annotated

`violation` / `outcome` を保持するか、名前が `*Result` / `*Outcome` / `*Report` / `*Verdict` /
`*Finding` で終わる公開 `struct` / `enum`（`#[cfg(test)]` より前の本文のみを対象に測定）：

| クレート | 型 |
|---|---|
| `fsl-runtime` (12) | `StepResult` `Violation` `BfsResult` `ExplicitResult` `ExplicitViolation` `ExplicitReachableWitness` `ReachableWitness` `RefinementCheck` `BoundedLivenessMonitor` `BoundaryProbe` `EnabledAction` `Monitor` |
| `fsl-verifier` (10) | `BmcResult` `BmcViolation` `InductionResult` `RankedLeadstoResult` `VacuityFinding` `ProgressCheck` `ImplicationResult` `ReachableWitness` `ReachableBlocker` `ReachableDiagnosis` |
| `fsl-tools` (6) | `DocumentCheckReport` `CausalError` `CausalWarning` `VariableRole` `Applicability` `SupportOverlay` |
| `fsl-syntax` (3) | `DomainAggregate` `DomainEffect` `DomainRetry` |
| `fsl-solver` (1) | `SatResult` |
| `fslc` (1) | `OutcomeClass` |

**うち `#[must_use]` が付いているもの：0 件。**

## Why this matters for a verifier

`AGENTS.md` は「記号検証・具体 Monitor・solver 非依存 BFS の三者が一致しなければならない。
**自信を持って緑になる false negative はクラッシュより危険**」と規定している。
検証結果や違反を運ぶ値が黙って捨てられる経路は、まさにその false negative を生む。

`docs/DESIGN-rust-component-internals.md` §2 hard gate 6 は
「Failures remain typed or structured until a delivery projection maps them to a stable JSON or
protocol contract. Unsupported shapes fail closed.」と述べている。
値が型として存在していても**破棄が検査されない**なら、この gate は静かに迂回される。

## Proposal

上記 33 型に `#[must_use]`（適切な場合は理由付きの `#[must_use = "..."]`）を付ける。

優先順位：

1. **`fsl-runtime` と `fsl-verifier`**（22 型）— 意味論の結果を直接運ぶ。#843 のクラス
2. **`fsl-solver`** の `SatResult` — solver の判定を捨てると偽陰性に直結する
3. **`fsl-tools`** / `fsl-syntax` / `fslc` の残り

`Monitor` と `EnabledAction` は結果型ではなく状態/記述子なので、
機械的に付けるのではなく個別に判断すべきである（本 issue の対象から外してもよい）。
その判断も本 issue のレビュー範囲に含める。

### 副次的な棚卸し（本 issue に含めるか分けるかは maintainer 判断）

`unreachable!()` 49 件、`panic!` 14 件。多くは `String` 型の演算子/名前に対する
`match` の catch-all（別 issue で扱う内部表現の話）に由来する。
`#[must_use]` とは独立だが、同じ「型が表現を許すが人間が到達不能と主張している」面である。

## Required controls

- **拒否対照（必須）**：`#[must_use]` を付けた型の値を `f()?;` の形で捨てるコードが
  `cargo clippy -- -D warnings` で**ビルド失敗する**ことを示すテスト。
  上記の最小再現がそのまま雛形になる。
  `rust/fslc/src/lib.rs:340` を修正する #843 の対応と組み合わせると、
  同型のコードが再導入できないことが機械的に保証される
- **受入対照**：正しく検査するコード（`let result = ...; if result.violation.is_some() { ... }`）が
  警告なしでビルドできること
- **回帰**：`cargo clippy --workspace --all-targets --locked -- -D warnings` が
  ワークスペース全体で通ること（既存の破棄箇所が他にあれば、それ自体が発見である）

**最後の点が重要である。** 33 型に付けた時点で clippy が落ちる箇所が出るなら、
それは #843 と同種の未発見の破棄であり、この issue の主目的はその列挙になる。

## Coupled changes

型注釈の追加のみであり、文法・lowering・意味論・公開 JSON 契約・exit code・location は変わらない。
`docs/DESIGN-rust-component-internals.md` の hard gate 7
（内部移動は公開契約や通常の Cargo 依存グラフを変えない）を満たす。

- `rust/fsl-runtime` / `fsl-verifier` / `fsl-solver` / `fsl-tools` / `fsl-syntax` / `fslc` の型定義
- `changelog.d/` フラグメント
- 破棄箇所が発見された場合はその修正と負の対照

`docs/LANGUAGE.md` / `docs/LANGUAGE.ja.md` / `skills/fsl/reference.md` は
言語機能の変更がないため対象外。

## Related

- **#843** — 本 issue の欠落が実際に通した欠陥。修正 PR に本 issue の
  `StepResult` 部分を含めるのが自然
- `docs/DESIGN-rust-component-internals.md` §2 hard gate 6（typed/structured failure）、hard gate 7

## Notes

調査は read-only で行っており作業ツリーへの変更はない
（`git status --short` は空、HEAD `abef68c`）。
測定は `#[cfg(test)]` より前の本文のみを対象にしており、テストコードは含まない。

rizumita:
Closes #843. Closes #858.

## Problem

`testgen`'s fixed-seed conformance walk (`testgen_trace_vectors`) is a concrete
Monitor run capped at 100 steps and **independent of `--depth`**. It discarded
`Monitor::step`'s `StepResult`, and because the Monitor rolls a violating step
back, the unchanged pre-step state was recorded as that step's `expected` — a
no-op semantics no FSL contract states.

Reproduced on the working tree (`rust/fslc/tests/fixtures/testgen_walk_late_violation.fsl`,
violation at step 30, default depth 8):

| command | before | after |
|---|---|---|
| `verify --depth 8` | `verified` / exit 0 | unchanged |
| `verify --depth 30` | `violated` `AtMost29` step 30 / exit 1 | unchanged |
| `testgen --target vitest` | **`generated` / exit 0** | **`violated` `AtMost29` step 30 / exit 1** |

The generated harness contained 100 walk entries of which ~71 were
`{"action":"inc","params":{},"expected":{"c":29}}` — asserting that `inc()`
leaves `c` at 29. What the spec states is that `inc()` is enabled at `c = 29`
(`requires c < 30`) and that its successor `c = 30` violates `AtMost29`.

Consequences, all measured:

1. **False positive** — a conforming implementation fails the generated test.
2. **False negative** — an implementation that silently does nothing at that
   point passes, and the invariant violation stays hidden behind a green suite.
3. **Target parity break** — `--target pytest` drives the walk live against the
   Monitor and `pytest.fail`s at step 30, while the five baked targets pass. Same
   spec, same seed, opposite conclusions. This is the dual-evaluator
   disagreement `AGENTS.md` ranks above everything else.

## Structural cause (#858)

`Result` is already `#[must_use]`, so

```rust
monitor.step(choice).map_err(|error| error.to_string())?;
```

satisfies the compiler: `?` discharges the `Err`, and the inner `StepResult` —
which owns `violation` — is dropped as a statement value. Annotating the
*function* cannot close this; it is redundant against `Result`. The attribute
has to be on the **type**.

26 public outcome-carrying types now carry it. Running
`cargo check --workspace --all-targets --locked --keep-going` with them in place
enumerated **every** discard in the tree:

| site | function | status |
|---|---|---|
| `rust/fslc/src/lib.rs:340` | `testgen_trace_vectors` | #843, reachable |
| `rust/fslc/src/main.rs:5332` | `requirement_trace_scenarios_from_source` | masked (see below) |
| `rust/fslc/src/main.rs:5366` | `requirement_trace_scenarios_from_source` | masked (see below) |
| 11 sites in `fsl-runtime`/`fsl-verifier` tests | — | discarded verdicts in tests |

All 14 are fixed. Annotating all 26 types surfaced **no production discard
beyond the 3 already known** — the enumeration is now complete rather than
assumed.

## Correction to the handoff: the `requirements` path is masked, not reachable

The handoff and #858 both state the defect "reaches the `requirements` scenario
path too". The compile-time measurement is right — those two sites do discard a
`StepResult`. The reachability claim is not: `run_scenarios_mode_from_source`
calls `validate_requirement_traces_from_source` at `main.rs:5102`, **before**
`requirement_trace_scenarios_from_source` at `main.rs:5284`, and
`validate_requirement_trace_source` already checks `result.violation.is_some()`
for both acceptance and forbidden-setup steps.

Measured on `rust/fslc/tests/fixtures/requirements_acceptance_walk_violation.fsl`:

```console
$ fslc scenarios requirements_acceptance_walk_violation.fsl
{"result":"error","kind":"acceptance","id":"AC-1","failed_step":1,"trace_type":"acceptance"}
$ echo $?
2
```

So those two guards are **defence in depth**, and their control is a
*preservation* control, not a rejecting one. They are not dead code: with the
upstream violation check disabled, and a spec that also clears the two other
masking layers (BMC at `--depth`, and validation's expectation check — the
rollback must leave the expectation true), the guard's own diagnostic is what
surfaces:

```
acceptance 'AC-1' step 9 violated invariant 'AtMost9' after validation accepted the trace
```

## Fix

`testgen_trace_vectors` now returns `TestgenWalk::{Clean, Violated}`.
`run_testgen` projects `Violated` onto `fsl_verifier::BmcResult` and renders it
through `render_bmc_output`, exactly as `render_explicit_output` projects the
solver-free explicit engine. That dispatches by violation kind the same way
`verify` does, so every envelope field matches `verify`'s for the same
violation. No harness is written.

**An earlier revision of this PR rendered straight through
`render_boundary_output` and was wrong.** That is `verify`'s
`partial_op`/`type_bound` path; for an `invariant`, `trans`, or `ensures`
violation it emits `loc`, `violating_bindings`, and `blame` as null, and reports
`trace_type` as the raw kind rather than `"sla"` for a generated `_deadline_`
property. Measured on the fixture in this PR:

| field | `render_boundary_output` (rejected) | `render_bmc_output` (shipped) | `verify` |
|---|---|---|---|
| `loc` | `null` | `{"line":14,"column":3}` | `{"line":14,"column":3}` |
| `violating_bindings` | `null` | `[{}]` | `[{}]` |
| `blame` | `null` | conjunct breakdown | conjunct breakdown |

That is a location and evidence loss, which `AGENTS.md` forbids allowlisting.
It was found by an independent soundness review, not by this PR's own tests —
which asserted `invariant` and `violated_at_step` but **not** `loc`, and
therefore passed the defect. The regression test now compares the **whole**
envelope against `verify` rather than a field list the implementer chose, and
that assertion is calibrated: reverting to `render_boundary_output` fails it
with `loc: testgen=null verify={"line":14,"column":3}`.

`domain testgen` inherits the refusal through `run_domain_testgen`'s verbatim
propagation, pinned by its own control with a real `domain` fixture.

Deliberately **not** an exit-2 `kind:"semantics"` error: `run_testgen`'s existing
comment (and #472) record that reclassifying "your spec has a bug" as "fix your
input" destroys the trace. The violating trace entry carries `attempted_state`,
so it records the successor the spec states (`c = 30`), not the rollback
(`c = 29`).

Not adopted: shortening the walk to `--depth` (#843 option B) would change every
committed golden and cut detection power; adding an `outcome` field to
`testgen-trace.v1` (option C) is a schema major across six emitters.

### A contract conflict this resolves explicitly

`docs/DESIGN-bridge.md` §3 item 3 prescribed the *opposite* resolution — emit the
harness and "distinguish it in the fail message as **a bug in the spec itself**".
That is sound only for `--target pytest`, whose walk runs live against the
Monitor at test time; the five baked targets embed a static fixture and have no
run-time oracle left to distinguish anything. Refusing uniformly across all six
rather than pytest-only is what #843's required target-parity control asks for —
one spec and seed must yield the same conclusion everywhere, and generating for
pytest while refusing the baked five would re-break parity in the other
direction. `AGENTS.md` requires resolving such a conflict rather than silently
picking a side, so the superseding decision and its rationale are recorded in
that design note.

The same note now records a three-way behavioral difference that was previously
undocumented: the frozen Python (`src/fslc/testgen.py`, `_bake_random_walk`)
never had this bug — it inspects the step result and breaks, baking the
violation-free prefix and exiting 0. So Python truncates silently, Rust-before
baked a no-op, and Rust-after refuses. No gate compares walk-violation behavior
across the two implementations, so nothing enforces either side; the table is the
record. The Rust workspace is authoritative and Python is a frozen reference, so
this is deliberate rather than a parity defect to allowlist.

## Evidence

Toolchain: build/test on 1.95.0, clippy on **1.98.0** (what CI resolves `stable`
to). Each worktree has its own private `CARGO_TARGET_DIR`.

**New controls** (`rust/fslc/tests/issue_843_testgen_walk_violation.rs`,
`issue_858_result_type_must_use.rs`) — 8 tests, all passing.

**Reversion experiment.** Each rejecting control was shown to fail with the fix
reverted in place, and each mutation was verified to have landed where aimed:

| reversion | result |
|---|---|
| R1 — discard the `StepResult` again in the walk | `a_walk_violation_past_the_verified_depth_aborts_generation` **FAILED**, `every_target_refuses_the_violating_walk_identically` **FAILED**; the 2 acceptance controls and the preservation control correctly still passed |
| R2 — remove `#[must_use]` from `StepResult` | all 3 of `issue_858`'s tests **FAILED** |
| R3a — remove both `requirements` guards | **no test failed** — confirming these are preservation controls. Their removal is caught by `clippy -D warnings` (two `unused StepResult` warnings), i.e. by #858's mechanism, not by a behavioural test |
| R3b — disable the upstream validation check | the guard's own diagnostic surfaces, proving it is live code |

**Golden preservation.** `all_six_public_kernel_targets_match_the_pre_migration_goldens`
and `compose_bridge_preserves_pytest_and_baked_target_goldens` pass: committed
harness bytes are unchanged for every target. All 5 existing `issue_472`
propagation contracts still pass.

**`--depth` independence** (unverified in #843, now measured). #843 states the
walk is "100 fixed". It is a *cap*: the walk also stops when no action is
enabled. `testgen_walk_clean.fsl` produces a 30-entry walk, byte-identical at
`--depth` 2, 4 and 12. The invariant `--depth` does not reach the walk is pinned
by `the_baked_walk_does_not_vary_with_depth`.

**Full gate**, all on 1.98.0 (what CI resolves `stable` to; the local default
is 1.95.0). `cargo fmt --check` clean; `cargo clippy --workspace --all-targets
--locked --keep-going -- -D warnings` clean, exit 0; `cargo test --workspace
--locked --no-fail-fast` — **222 suites, 1522 passed, 0 failed**.

### Findings from independent review, all fixed here

Two read-only reviewer agents audited the PR head. Between them and the test
suite they found five defects this PR's own evidence had missed:

| # | Finding | Severity |
|---|---|---|
| 1 | `render_boundary_output` dropped `loc`/`violating_bindings`/`blame` vs `verify` | soundness-adjacent contract drift |
| 2 | `docs/DESIGN-bridge.md` prescribed the opposite resolution — a real contract conflict, resolved silently | contract drift |
| 3 | `docs/DESIGN-kernel-contract.md:212,:375` described the old signature and behavior | contract drift |
| 4 | the three new `.fsl` fixtures lacked the Apache-2.0 SPDX header | coupled-change gap |
| 5 | six line citations into `main.rs`/`outcome.rs` went stale, the same class as the `mutants.toml` pins | coupled-change gap |

A sixth was caught by the suite rather than by any reviewer:
`rust/fsl-core/tests/domain_render_agreement.rs`'s
`corpus_discovery_matches_registered_classification` enumerates every *domain*
declaration under `examples/` and `rust/fslc/tests/fixtures`, so the new domain
fixture had to be registered. It is registered in `VALID_DOMAIN_FIXTURES` — not
the `KNOWN_DIVERGENT` allowlist — and satisfies
`valid_domain_corpus_lowering_and_rendering_agree`.

**Corpus sweep.** `testgen` over every `.fsl` under `specs/` and `examples/`,
comparing each spec's `scenarios` exit code with its `testgen` exit code, so
specs this change newly rejects are separated from specs already rejected:

| | count |
|---|---|
| cleanly measured | 212 |
| **newly rejected by this change** | **0** |
| unchanged, still generates (`0,0`) | 132 |
| already non-zero before (`1,1` / `2,2`) | 72 |
| `scenarios` 1 / `testgen` 0 | 8 |
| unmeasured (hit a 120 s watchdog in *both* commands) | 2 |

The 8 asymmetric rows are the documented `--strict` distinction, not a
regression: `testgen` passes `allow_unreached = !strict`, so non-strict
generation tolerates an unwitnessed reachable that bare `scenarios` rejects.
Verified on `examples/nfr/sla_worker.fsl` — `scenarios` gives
`reachable_failed` with 1 unreached, and `testgen --strict` reproduces it
exactly.

The 2 unmeasured specs (`examples/agentic_rag/agentic_rag_design.fsl`,
`agentic_rag_requirements.fsl`) exceeded 120 s in `scenarios` *alone*, which
runs entirely before this change's code path, so the cost is upstream of it.
They are recorded as unmeasured rather than as passes.

## Coupled changes

No grammar, lowering, typed-model, JSON envelope, exit-code, or location change,
so no LSP index change is required. `docs/DESIGN-rust-component-internals.md`
hard gate 7 (internal moves do not change public contracts) holds.

- `docs/LANGUAGE.md` + `docs/LANGUAGE.ja.md` — the walk's `--depth` independence
  and its fail-closed behaviour, section-aligned (18/18 `## ` sections;
  `tools/build_site_reference.py` clean)
- `skills/fsl/reference.md` §9 — same statement for agent authors
- `docs/intro/language.{en,ja}.html` — regenerated by
  `tools/build_site_reference.py` (required by
  `.github/workflows/site-reference-freshness.yml`); the diff is only the new
  paragraph in each language
- `rust/.cargo/mutants.toml` — the three line-scoped mutation exclusions that
  file documents as "coupled to unique source anchors by
  `implementation_mutation_manifest.rs`" move from 420/427/451 to 425/432/456,
  the five `#[must_use]` lines inserted above them in `fsl-verifier/src/bmc.rs`.
  The anchor strings are the authority and are computed live from source by that
  test; the regex list is the derived side. Caught by
  `p2_critical_scope_and_equivalence_manifests_are_current`, which is exactly
  what it is for
- `changelog.d/843-…fixed.md`, `changelog.d/858-…fixed.md`

## Accounting for all 33 types in #858

#858's 33 are **(crate, type) pairs**, and `ReachableWitness` is declared in both
`fsl-runtime` and `fsl-verifier`. So the 25 distinct names this PR annotates are
**26 declarations**, and 33 − 26 = **7** excluded, not 8. (The diff shows 27
`#[must_use]` additions; the 27th is `TestgenWalk`, a new type this PR
introduces, not one of the 33.)

| # | Type | Crate | Why no `#[must_use]` |
|---|---|---|---|
| 1 | `Monitor` | `fsl-runtime` | stateful interpreter held across steps; its verdicts leave through `StepResult`, which is annotated |
| 2 | `EnabledAction` | `fsl-runtime` | descriptor produced by enumeration — an input to `step`, not a verdict |
| 3 | `BoundedLivenessMonitor` | `fsl-runtime` | stateful tracker; all five fields private, carries a `next_step` cursor across observations. Same category as `Monitor`: verdicts leave through `observe() -> Result<Option<BoundedLivenessViolation>, _>` and `status() -> BoundedLivenessStatus` |
| 4 | `VariableRole` | `fsl-tools` | `Copy` classification enum labelling a causal variable. One of its four variants is literally named `Outcome`, which is why a grep for `outcome` selected it |
| 5 | `DomainEffect` | `fsl-syntax` | surface AST node. Its `pub outcomes: Vec<String>` is a list of *event names* for effect outcome roles — the same grep artifact, not a verdict |
| 6 | `DomainAggregate` | `fsl-syntax` | surface AST node. Every field checked: no `outcome`/`violation` field at all, so #858's stated rule does not actually select it |
| 7 | `DomainRetry` | `fsl-syntax` | surface AST node, derives `Default`; no outcome field either |

None of the seven is an oversight. Each is recorded in `EXCLUDED` in
`rust/fslc/tests/issue_858_result_type_must_use.rs`, which fails if one of them
gains the attribute or if an annotated type loses it.

### A limitation #858 does not state: `#[must_use]` does not reach inside `Option<T>`

#858's insight is that `Result` being `#[must_use]` is satisfied by `?` while the
inner value escapes. The same is true one level deeper, and it is worse:
`Option<T>` is **not** `#[must_use]` in `core`, and a `#[must_use]` on `T` does
**not** propagate through it. Compiler-verified on 1.98.0:

```rust
#[must_use] pub struct Verdict { pub bad: bool }
fn f() -> Result<Option<Verdict>, String> { Ok(None) }
fn g() -> Result<Verdict, String> { Ok(Verdict { bad: false }) }
// f()?;  -> no warning
// g()?;  -> warning: unused `Verdict` that must be used
```

So any `-> Result<Option<Outcome>, E>` signature is unprotected no matter how the
outcome type is annotated. `BoundedLivenessMonitor::observe`
(`rust/fsl-runtime/src/lib.rs:1180`, `:1188`) has exactly that shape and carries
a liveness violation. `BoundedLivenessViolation` is not among #858's 33, and the
census rule in this PR does not select it either — it has no
`*Result`/`*Outcome`/`*Report`/`*Verdict`/`*Finding` suffix and no `violation:`
field, which means **the rule misses every `*Violation` type by name**.

Current status, verified: all four production call sites inspect the result
(`rust/fslc/src/main.rs:4221`, `:4358`, `rust/fslc/src/causal.rs:852`, `:1136`).
Six test sites drop it (`rust/fsl-runtime/tests/bounded_liveness.rs:71`, `:134`,
`:155`, `:157`, `:175`, `:177`) — the same "test ignoring the Monitor's verdict"
pattern this PR converted 11 of, except that no lint can ever flag these. There
is no live production defect, and annotating `BoundedLivenessViolation` would be
**ineffective** rather than merely unnecessary, so it is filed as a follow-up
rather than changed here.

## Exclusions from #858, with reasons

`Monitor`, `BoundedLivenessMonitor` (stateful interpreters/trackers),
`EnabledAction` (an enumeration descriptor), `VariableRole` (a classification),
`DomainAggregate`/`DomainEffect`/`DomainRetry` (surface AST nodes). Recorded in
`EXCLUDED` in `issue_858_result_type_must_use.rs`, which fails if one of them
gains the attribute or an annotated type loses it.

**Note on #858's count.** The issue reports 33 types from a hand-curated table.
The mechanical rule as the issue states it (`*Result`/`*Outcome`/`*Report`/
`*Verdict`/`*Finding`, or a `violation`/`outcome` field) reproduces **12** of
them. Both numbers are pinned: `DISCOVERED` (12) is the drift detector for newly
added result types, `ANNOTATED` (25 names / 26 declarations) is the reviewed
decision.


rizumita:
## Independent review — one Medium, two Lows. Not merging.

Reviewed by an agent that did not implement this PR. The Medium lands on the PR's own central claim and is
confirmed by mutation, so it blocks the merge.

### Medium — `assert_envelope_parity` is not a whole-envelope comparison

`rust/fslc/tests/issue_843_testgen_walk_violation.rs:164` excludes `cache`, `versions`, and eight
"run-dependent" fields — including **`depth`, `engine`, and `statistics`** — and `:209` filters them
before comparing. Both parity tests call the helper.

**CONFIRMED:** changing `main.rs:10865` from `depth: violation.step` to `depth: 0` left **all 7 tests
passing**. The value is genuinely exposed, through `verification_output.rs:1737`, so a late invariant walk
would report `depth: 0` while `verify --depth 30` reports `depth: 30` — precisely the envelope divergence
this control exists to reject.

The helper's own doc comment says:

> Two revisions of this fix diverged from `verify` in OPPOSITE directions and both passed a hand-picked
> field list. ... Only a full comparison rejects both

It is still a hand-picked field list. The picking moved from an allowlist to a denylist. That is worse than
an ordinary gap because the comment tells the next reader the hole is closed — and I repeated that claim in
my own reporting, so I propagated it rather than catching it.

Required before this merges: split the exclusion list by *reason*, not by the label "run-dependent". Cache
presence is genuinely ambient and its subset check is right — that was established empirically. `depth` is
a contract field `verify` and `testgen` must agree on. Contract fields compare exactly; anything excluded
must state **why it cannot be compared**. Then re-run the `depth: 0` mutation and show it failing.

### Low — changelog fragments omit the rendered-bullet lead

`changelog.d/843-testgen-walk-violation.fixed.md` must begin `Fixed (#843): ` and
`changelog.d/858-result-type-must-use.fixed.md` must begin `Fixed (#858): `. The aggregator does not
invent the prefix.

### Low — stale counts inside the census control

`issue_858_result_type_must_use.rs:24`, `:62`, `:91` cite 26/25/26/12 against actual 27 annotated names,
28 declarations, 16 discovered names. Stale numbers inside a census control describe something other than
what the control checks.

### What held, each confirmed rather than assumed

- **Renderer routing is correct** by direct source inspection: the concrete pre-pass is limited to
  `partial_op`/`type_bound`; the Monitor's remaining kinds are `invariant`, `trans`, `ensures`, all
  through `render_violation`. This was my top suspicion and it holds.
- **The refusal detector is real.** Temporarily discarding the walk violation — reproducing #843 itself —
  failed 4/7 including the six-target refusal test, while the clean-walk and requirements preservation
  tests stayed green, matching their labels. A detector and a preservation control behaving differently
  under the same mutation is the distinction worth having.
- All 11 changed `StepResult` sites assert a semantic result rather than binding and discarding.
- The mutant line pins point at the documented BMC anchors; `429`->`428` failed
  `p2_critical_scope_and_equivalence_manifests_are_current` with the expected/actual exclusion sets.
- The `Option` limitation is accurately stated in the commit body and both changelog prose, with no
  blanket impossibility claim remaining.

### Stated limits

Workspace-wide Clippy and tests were not run — two cargo tracks is this machine's measured ceiling and both
were occupied; CI covers those (18 SUCCESS / 3 SKIPPED). A separate late `partial_op` fixture was not
executed; its route was verified from the runtime and renderer source instead.

Follow-ups filed from p1's own findings while this was in review: #868 (`#[must_use]` does not propagate
through `Option`), #872 (Rust/Python `testgen` walk-violation divergence, framed as a contract gap rather
than a code defect), #873 (`mutants.toml` hand-pinned line numbers against values the test computes live).

rizumita:
## Correction to my previous comment — the Medium's mechanism was wrong, and the underlying problem is different

I posted the review's Medium as CONFIRMED, including the claim that `depth` is exposed through
`verification_output.rs:1737` so a late invariant walk would report `depth: 0` while `verify --depth 30`
reports `depth: 30`. **That is false**, and I should have followed the citation before repeating it.

`:1737` is inside `add_common`, and `add_common` is called from exactly four places —
`render_explicit_budget` (`:1215`), `render_explicit_success` (`:1274`), `render_reachable_failure`
(`:1537`), and `render_success` (`:1673`). **Neither renderer on the violation path calls it.** The full
key set each violation renderer emits:

    render_violation:        blame invariant origin result spec trace trans
                             violated_at_step violating_bindings violation_kind
    render_boundary_output:  blame origin result spec trace_type trace
                             violated_at_step violating_bindings violation_kind

No `depth`, no `engine`, no `statistics`, no `closure`, no `states_explored`, no
`max_frontier_width`, no `elapsed_s`. So the `depth: 0` mutation left 7 tests passing because **the
field is not in the violation envelope at all**, not because an exclusion list hid it. The reviewer's
citation did not verify its own claim — the same defect class this repository keeps finding, this time in
a review rather than in code.

**The real problem is worse than a too-wide exclusion list, and the implementer diagnosed it correctly
while I was repeating the wrong version.** From the reworked doc comment:

> an earlier revision said so while excluding eight "run-dependent" fields, of which SEVEN appear in
> NEITHER envelope on this path — the list was copied from `BmcOutputOptions`' field names, not read off
> the JSON, so it excluded nothing and merely looked considered

An exclusion list assembled from a Rust struct's field names rather than from the actual JSON. Seven of
eight names were dead. The list did not weaken the comparison; it created the appearance of a considered
one. `cost` is the only genuine exclusion, and it now carries the reason it *cannot* be compared — the
walk runs no solver, so `testgen`'s block is structurally zero — rather than a category label.

The structural fix matters more than the wording: an `EXCLUSIONS_ARE_LIVE` check now fails if an
exclusion names a key absent from both envelopes, so a dead exclusion cannot sit there looking deliberate
again. That closes the class, where editing the list would only have closed the instance.

### What stands from the review

The Medium was right that the doc comment overclaimed, and right that this blocks the merge until the
comparison says what it does. Both Lows stand unchanged — the missing `Fixed (#nnn): ` changelog leads
and the stale 26/25/26/12 counts against actual 27/28/16.

### What I got wrong, stated for the record

Three times today I asserted a mechanism from reading and was wrong: the `failed_link` "defect" in #850
(retracted), the loop-status explanation in #867 (#871), and now this. Each was settled in seconds by
following a citation or running a two-line probe. The one thing that has reliably worked is checking
whether the cited evidence verifies the claim — which is how this correction got found.

rizumita:
## Merge conditions met; pushed the real merge candidate

CI had only ever seen `995886a`, which predates all of the review fixes. The head is now `f83d0f91` —
`68eb308` rebased onto `382c2b4` — and this is the first time CI evaluates it.

### The Medium is closed, and the fix is structural rather than a list edit

`assert_exclusions_are_live` is now an extracted, named function, and its assertion says what to do about
the failure rather than only that one occurred:

> exclusion '{key}' names a field one of the envelopes does not carry, so it excludes nothing: delete it
> from NOT_COMPARABLE rather than leaving it to imply a considered decision

Calibrated by the implementer against the committed SHA and **reproduced independently by me** on the
rebased head:

| | |
|---|---|
| dead exclusion `"engine"` added | `5 passed; 2 failed` — both violation tests, each printing the message above |
| removed again | `7 passed; 0 failed` |

**Both violation tests fail, not just one**, because the check lives in the shared helper and so fires from
every caller. That is the difference between fixing the instance and closing the class: editing the
exclusion list would have removed the seven dead entries, while the liveness check makes a dead entry
impossible to add back silently.

`NOT_COMPARABLE` is down to `["cost"]`, with the reason stated as why it *cannot* be compared — the walk
runs no solver, so `testgen`'s block is structurally zero while `verify`'s reports real solving, which are
different quantities rather than the same quantity within a tolerance. `VERIFY_ONLY` keeps `cache` as a
subset check because its **presence** is ambient, with the cold-cache exemption preserved.

### Both Lows closed

`Fixed (#843): ` and `Fixed (#858): ` leads present. Counts corrected and cross-checked prose against code
per item: "The 27 public type names" / `ANNOTATED: [&str; 27]`; "27 distinct names but 28 declarations" /
`ANNOTATED_DECLARATIONS = 28`; "selects 16" / `DISCOVERED: [&str; 16]`. Stale 26/25/12 occurrences: zero.

### Verified by me on `f83d0f91`

`issue_843_testgen_walk_violation` 7 passed, `issue_858_result_type_must_use` 3 passed, and the
dead-exclusion calibration above in both directions with the file restored afterwards.

Reported by the implementer against `68eb308` and not re-run by me: 222 suites, 1523 passed, 0 failed;
R1 and R5 re-run after the extraction to show it did not weaken the real detectors; fmt and Clippy 1.98.0
clean. Those numbers are from one run against that SHA, which is the rule we settled on after a report
mixed a current clippy result with a stale suite figure.

### Correction that stands on the record

The review's Medium named `depth` as an excluded contract field exposed via `verification_output.rs:1737`.
That citation does not support the claim — `:1737` is inside `add_common`, which no violation renderer
calls — and I repeated it here before checking. The real problem was worse and the implementer diagnosed
it: seven of eight exclusions named fields absent from **both** envelopes, transcribed from
`BmcOutputOptions`' field names rather than read off the JSON. The list weakened nothing and manufactured
the appearance of a considered decision.

Follow-ups from this PR's own findings, all filed: #868, #872, #873. #868 led to #875, a live false
`pass` on a liveness violation, fixed and merged as #876.

Issue #865, opened 2026-08-21
Add behavioral controls for the release GLIBC ABI guard


rizumita:
## Finding

The Linux release ABI step enforces a non-empty GLIBC requirement and a maximum of GLIBC_2.39, but its only existing repository control is a static workflow-text assertion. There is no accepting/rejecting behavioral calibration for those GLIBC checks.

## Scope

Add a pre-merge behavioral control for the existing GLIBC guard. Do not change the threshold or release artifact contract as part of this finding.

Discovered while fixing #861; tracked separately to keep the libz3 linkage fix focused.

PR #867, opened 2026-08-22
fix(release): make the libz3 linkage guard actually fail the release job

### LINKED ISSUE #861: release.yml の libz3 動的リンク検査が errexit を回避しており、破っても job が落ちない
# Problem

`.github/workflows/release.yml:92` のリリース成果物検査で、**`libz3` の動的リンクを
検出しても job が落ちない**。`!` を先頭に付けた行は `errexit` の対象外になるため。

`actionlint` が SC2251 として報告する:

```
.github/workflows/release.yml:92:9: shellcheck reported issue in this script:
  SC2251:info:5:3: This ! is not on a condition and skips errexit.
  Use `&& exit 1` instead, or make sure $? is checked [shellcheck]
```

# 該当箇所

```yaml
        run: |
          for binary in rust/target/release/fslc rust/target/release/fslc-lsp; do
            required=$(readelf --version-info "$binary" | grep -o 'GLIBC_[0-9.]*' | sort -V | tail -1)
            test -n "$required"
            test "$(printf '%s\n' GLIBC_2.39 "$required" | sort -V | tail -1)" = GLIBC_2.39
            ! ldd "$binary" | grep -q libz3
          done
```

同じブロック内の `test -n` と `test ... = GLIBC_2.39` は `errexit` が効くので、
失敗すれば job が落ちる。**`! ldd ... | grep -q libz3` の行だけが効かない。**

# なぜスタイル指摘ではないか

この行が検査しているのは「リリースバイナリが `libz3` を動的リンクしていないこと」で
ある。`fsl-runtime` が solver から独立していること、および配布物が Z3 共有ライブラリを
必要としないことは、この repository の中心的な不変条件に属する
（AGENTS.md: `fsl-runtime` must remain independent of `fsl-solver`, Z3, and
JavaScript solver bridges）。

つまり **配布物が libz3 に依存してしまっても、リリースワークフローは緑のまま通る。**
検査は書かれているが、破っても落ちない。今日この repository で6件修正したのと同じ
「何も強制しない制御」である。

# 未実測

- 実際に `libz3` を動的リンクしたバイナリでこのワークフローを走らせて、緑のまま通ることを
  観測していない。上記は `set -e` と `!` のシェル意味論からの推論と、shellcheck の
  診断に基づく。**着手時に実測で確認すること。**
- `bash` の `run:` が実際に `-e` 付きで起動されているかは
  `.github/workflows/release.yml` の該当 job の shell 設定を確認すること
  （GitHub の既定は `bash -e {0}`）。

# 期待される挙動

1. `! cmd` を errexit の効く形に書き換える。shellcheck の助言どおり
   `ldd "$binary" | grep -q libz3 && exit 1`、または `if ldd ... | grep -q libz3; then exit 1; fi`。
2. **rejecting control**: `libz3` に動的リンクされたバイナリ（またはそれを模した
   fixture）でこの検査が**落ちる**こと。現状これを確認する対照が無い。
   正のパスだけが緑なのは、この検査がドリフトを検出できることの証明にならない。
3. 同じブロックの GLIBC 検査にも accepting/rejecting 対照があるか確認する。
   無ければ併せて与える。

# 発見経緯

PR #854（toolchain 固定）のレビュー中に `actionlint` が報告した。#854 は
`release.yml` を変更していない（diff 0件）ため、そちらには含めなかった。
レビュアーの判断も「pre-existing かつ無関係、#854 の blocker ではない、
追跡が無ければ別 issue に」であり、それに従う。

# 影響範囲

`.github/workflows/release.yml`

rizumita:
## Problem

`release.yml`'s Linux ABI step guarded against a dynamically linked Z3 with a bare `!` command inside
a loop:

```yaml
for binary in rust/target/release/fslc rust/target/release/fslc-lsp; do
  required=$(readelf --version-info "$binary" | grep -o 'GLIBC_[0-9.]*' | sort -V | tail -1)
  test -n "$required"
  test "$(printf '%s\n' GLIBC_2.39 "$required" | sort -V | tail -1)" = GLIBC_2.39
  ! ldd "$binary" | grep -q libz3          # <- never failed the job
done
```

The two `test` lines above it do fail the step. That last line did not. With `set -e`, a failing
non-final loop iteration does not abort, and the loop's exit status is the status of the **last**
iteration — so a `libz3`-linked `fslc` followed by a clean `fslc-lsp` passed. This is also why the
form looks correct under casual reading and would have behaved correctly with a single binary: the
defect only appears once the loop has more than one iteration.

The guard exists because a release binary that dynamically links `libz3` fails on a machine without a
matching Z3, which is the failure mode users hit first and maintainers see last.

## Contract change

The check moves out of the workflow into `tools/check-release-binary-linkage.sh`, invoked with the
same two binaries the loop covered — `rust/target/release/fslc` and `rust/target/release/fslc-lsp`,
verified to be exactly the loop's set, so coverage does not shrink. The script runs under
`set -euo pipefail` and reports the offending `ldd` lines rather than only failing:

```bash
if linked="$(grep libz3 <<<"$ldd_output")"; then
  printf 'release linkage check failed: %s dynamically links libz3:\n%s\n' "$binary" "$linked" >&2
  return 1
fi
```

No `!`, so nothing is exempt from `errexit`, and each binary is checked on its own status.

`--self-test` is wired into `check_rust_checks()` in `tools/check-native-integration.sh`, which is
already invoked from the required lane. `.github/workflows/ci.yml` is deliberately untouched: #863
changes that file, and the lane needs no new wiring there.

## Test evidence

`./tools/check-release-binary-linkage.sh --self-test` — exit 0, printing all three edges:

```
release linkage self-test: unfixed guard accepts a libz3-linked non-final binary
release linkage self-test: clean ldd fixture accepted
release linkage self-test: libz3 ldd fixture rejected
```

The first line is the calibration that matters: the self-test reconstructs the **pre-fix** loop and
asserts it accepts a `libz3`-linked non-final binary. If that reconstruction ever stops reproducing
the hole, the self-test fails rather than quietly passing — so the control is pinned to the defect it
was written for, not merely to the fixed behaviour.

Re-run by me after rebasing onto `c92b232`: exit 0, same three lines.

Also: `actionlint .github/workflows/release.yml` exit 0 with SC2251 gone and no other findings;
`./tools/aggregate_changelog.sh check` exit 0.

## Stated limits

The self-test installs a PATH stub `ldd`. It assumes a real Z3 dependency appears on an `ldd` output
line containing the literal substring `libz3`. **Real Linux `ldd` output was not verified locally**
(this is macOS); the stub stands in for it and does not claim to inspect a real release artifact. The
first release build after this merge is the first observation of the real thing.

The `GLIBC_2.39` assertions immediately above have only a static presence check and no behavioural
control — a positive path with nothing proving it can reject a too-new binary. That is out of scope
here and tracked in #865.

Closes #861

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,8 +94,8 @@ jobs:
             required=$(readelf --version-info "$binary" | grep -o 'GLIBC_[0-9.]*' | sort -V | tail -1)
             test -n "$required"
             test "$(printf '%s\n' GLIBC_2.39 "$required" | sort -V | tail -1)" = GLIBC_2.39
-            ! ldd "$binary" | grep -q libz3
           done
+          ./tools/check-release-binary-linkage.sh rust/target/release/fslc rust/target/release/fslc-lsp
 
       - name: Smoke test (Windows)
         if: runner.os == 'Windows'
--- a/changelog.d/861-release-libz3-abi-guard.fixed.md
+++ b/changelog.d/861-release-libz3-abi-guard.fixed.md
@@ -1,0 +1,1 @@
+Fixed (#861): make the release Linux linkage guard reject binaries dynamically linked to libz3, with accepting and rejecting ldd fixtures. The separate GLIBC behavioral-control finding is tracked in #865.
--- a/tools/check-native-integration.sh
+++ b/tools/check-native-integration.sh
@@ -38,6 +38,7 @@ check_rust() {
 # using `cargo-nextest`, which nextest cannot do for doctests. Doctests
 # therefore stay here, explicit and unsharded, rather than silently dropped.
 check_rust_checks() {
+  ./tools/check-release-binary-linkage.sh --self-test
   check_stack_parity
   cargo fmt --manifest-path rust/Cargo.toml --all -- --check
   cargo clippy --manifest-path rust/Cargo.toml --workspace --all-targets --locked -- -D warnings
--- a/tools/check-release-binary-linkage.sh
+++ b/tools/check-release-binary-linkage.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+check_binary() {
+  local binary="$1" ldd_output linked
+
+  ldd_output="$(ldd "$binary")"
+  if linked="$(grep libz3 <<<"$ldd_output")"; then
+    printf 'release linkage check failed: %s dynamically links libz3:\n%s\n' "$binary" "$linked" >&2
+    return 1
+  fi
+}
+
+run_check() {
+  local binary
+  for binary in "$@"; do
+    check_binary "$binary"
+  done
+}
+
+run_self_test() {
+  local mode="$1" fixture unfixed output
+  fixture="$(mktemp -d)"
+  trap 'rm -rf "$fixture"' RETURN
+  mkdir -p "$fixture/bin"
+  : >"$fixture/fslc"
+  : >"$fixture/fslc-lsp"
+  : >"$fixture/fslc-with-libz3"
+
+  # This PATH stub stands in for Linux ldd so the shell control can run on
+  # any host; it does not claim to inspect a real release artifact.
+  printf '%s\n' \
+    '#!/usr/bin/env bash' \
+    "case \"\$1\" in" \
+    '  *with-libz3) printf "\\tlibz3.so.4 => /fixture/libz3.so.4 (0x00000000)\\n" ;;' \
+    '  *) printf "\\tlibc.so.6 => /fixture/libc.so.6 (0x00000000)\\n" ;;' \
+    'esac' \
+    >"$fixture/bin/ldd"
+  chmod +x "$fixture/bin/ldd"
+
+  # Calibration: this is the pre-fix loop. A bad non-final binary is accepted
+  # because the final clean iteration supplies the loop's successful status.
+  unfixed="$fixture/unfixed-guard.sh"
+  printf '%s\n' \
+    '#!/usr/bin/env bash' \
+    'set -e' \
+    'for binary in "$@"; do' \
+    "  ! ldd \"\$binary\" | grep -q libz3" \
+    'done' \
+    >"$unfixed"
+  chmod +x "$unfixed"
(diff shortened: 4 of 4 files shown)


rizumita:
## Retrospective review finding — my mechanism explanation in this PR body is wrong

I merged this without an independent review, which was my mistake. The retrospective review found one
P2, and it is against my own writing rather than the code.

This PR body and `46ce2db`'s commit body say the defect was that the loop returns the last iteration's
status. That is only half the cause. Confirmed by execution, not by reading:

| reproduction | result |
|---|---|
| the pre-fix loop `! ldd "$binary" | grep -q libz3`, dirty then clean binary | **exit 0**, loop completed |
| the same loop with a bare `test` failing on a non-final iteration | **exit 1**, aborted before completion |

The second row settles it. Under `set -e` a bare non-final failure does **not** reach the end of the
loop, so final-iteration status cannot by itself explain the hole. The `!` prefix is the necessary
condition — it suppresses the abort — and only then does the loop's last status mask the earlier
failure. Both are required.

The provenance matters for anyone reading the history: the implementer's original framing was `!`
exempting the line from `errexit`, which was correct. I judged the loop-status framing more precise,
directed it into the commit message, and asserted that from reading rather than from the two-line shell
probe that would have settled it. Tracked in #871 along with the `tools/check-release-binary-linkage.sh:43`
comment, which carries the same wording in live code documentation.

Confirmed non-findings from the same review, each executed read-only:

- the self-test faithfully reconstructs the prior vulnerable command at parent `release.yml:93-98`,
  including `set -e` and the `!` pipeline — so the calibration pins the real defect, not a hypothetical one
- coverage is preserved: old and new checks cover exactly `fslc` and `fslc-lsp` on both Linux matrix
  legs, and no Windows/macOS binary previously received the `ldd` check
- the stub uses ordinary GNU `ldd` resolved-library shape, and the new `grep libz3` also matches an
  unresolved `libz3.so… => not found` line. Plausible from output-format analysis, not real-Linux
  execution — which this PR already disclosed
- #865's gap is real: `rust/fslc/tests/native_integration.rs:423` only asserts the workflow text
  contains `GLIBC_2.39`, with no behavioural calibration. Its `test` commands do **not** share the old
  loop-status defect
- the self-test is invoked at `tools/check-native-integration.sh:41`, and the required `rust checks` job
  reaches it: this PR's own paths produce `run=true, reason=product-paths-changed`

No behavioural defect in the merged guard. It rejects a `libz3`-linked binary as intended.

PR #869, opened 2026-08-22
fix(core): explain when a const shadows an enum member used as a map key

### LINKED ISSUE #855: const が enum メンバと同名だとそのメンバをマップキーに書けないが、診断は型不一致しか述べずシャドーイングを述べない
# Problem

`const` が `enum` メンバと同名のとき、その名前は `const` に解決される。結果として
そのメンバは**マップキーとして書けなくなる**が、診断はその原因を述べない。
作者から見て正しく見えるスペックが、型不一致だけを理由に落ちる。

# 実測（2026-08-21）

```fsl
spec EnumConstName {
  const A = 0
  enum Key { A, B }

  state {
    m: Map<Key, Bool>
  }

  init {
    m[A] = true
    m[B] = false
  }

  action noop() { }
}
```

`fslc check`:

```json
{
  "result": "error",
  "kind": "semantics",
  "message": "invalid init statement: expression of type Int is not assignable to Named(\"Key\") at 12:5",
  "loc": { "line": 12, "column": 5 }
}
```

`m[B]` は問題なく、落ちるのは `m[A]` の側だけである。つまり `A` が
`const A = 0`（`Int`）に解決され、`Key::A` には解決されていない。

# これは健全性の欠陥ではない

**沈黙も誤った意味論の採用も起きていない。** `check` は位置付きの
`kind: semantics` で明確に落ちる。`Int` を enum キーとして受理してしまう経路は無い。
本 issue は診断品質の話であり、優先度は低い。

（このことは #825 の作業中に、`rust/fsl-runtime/src/explicit.rs` の
`assignment_coverage` で整数 `const` キーを解決する際の型ゲートを検証していて判明した。
`resolve_enum_key` を先に試し `resolve_integer_const_key` を後にする優先順は、
仮にその順序が逆転しても型検査がここで弾くため、二重に守られている。）

# 期待される挙動

どちらかを選ぶ:

1. **診断を原因に即したものにする。** キーの位置に現れた識別子が、そのマップの
   キー型が属する enum のメンバ名と一致し、かつ同名の `const` に解決された場合、
   そのシャドーイングを述べる。例:
   `'A' resolves to const A = 0 (Int), not to member A of enum Key; a const shadows
   an enum member of the same name`。
   現在のメッセージは症状（型不一致）だけを述べ、原因（シャドーイング）を述べない。
2. **宣言時点で拒否する。** `const` 名が既存 `enum` のメンバ名と衝突する宣言自体を
   `check` で落とす。こちらの方が強いが、既存コーパスを壊す可能性があるので
   影響調査が必要。

いずれの案でも、accepting/rejecting 対照を伴わせること:
衝突がある場合に落ち、無関係な同名でない `const` と `enum` が共存する場合は通ること。

# 既存の対照

`rust/fslc/tests/explicit_engine.rs` の
`enum_map_key_name_collision_with_an_integer_const_is_rejected_before_runtime`
が現在の診断文を厳密に固定している（#825 で追加）。
本 issue を実装する際はその期待値を更新することになる。

# 影響範囲

`rust/fsl-core`（init 文の型検査と名前解決）、
`rust/fslc/tests/explicit_engine.rs`

rizumita:
## Problem

When a `const` shares a name with a member of a map's key enum, the name resolves to the `const`, so
that member becomes unwritable as a map key. The diagnostic reported only the type mismatch:

```
invalid init statement: expression of type Int is not assignable to Named("Key") at 12:5
```

`m[B] = false` is fine; only `m[A] = true` fails, because `A` resolved to `const A = 0` rather than
`Key::A`. From the author's side the spec looks correct and the message names neither the shadowing nor
what to do about it.

## What this is not

This is **not** a soundness defect and the fix does not treat it as one. Nothing is silently accepted
and no wrong semantics is adopted: `check` fails with a located `kind: semantics` error, and there is no
path that accepts an `Int` as an enum key. Name-resolution precedence is unchanged — altering it would
be a language contract change with a far larger blast radius than the problem, and it is not what #855
asks for.

## Contract change

The existing message, error kind, exit code, and location are all preserved. The diagnostic gains the
cause and a hint:

```
invalid init statement: expression of type Int is not assignable to Named("Key"); cause: map key 'A'
resolves to const A = 0 (Int), not to member A of enum Key because a const shadows an enum member of
the same name; hint: rename the const or enum member to make the map key unambiguous at 12:5
```

**The hint rides inside `message` because the semantic-error envelope has no `hint` field.**
`render_semantic_error` (`rust/fslc/src/verification_output.rs:137`) takes `message`, `loc`, and
`name_resolution` only; the `hint` key at `main.rs:13419` is a pass-through for envelope shapes that
have one. Adding a hint channel to semantic errors would be an envelope contract change outside this
issue's scope. Noting it explicitly so a reviewer does not read the embedding as a style choice and
"fix" it into a field that does not exist.

Related observation, not addressed here: `docs/LANGUAGE.md` specifies check-time rejection "with a hint"
in places, and **#841** is an open three-way mismatch about exactly that. If the documented contract
requires a hint on a semantic error, this missing channel is part of #841's scope.

## Controls

The rejecting control is the pre-existing test for this case, whose expected message is extended and
which now also pins the location (`assert_eq!(result["loc"], json!({"line": 12, "column": 5}))`). Two
accepting controls are added, and the pair is what matters:

| fixture | contents | shows |
|---|---|---|
| `issue_855_enum_map_key_without_const_collision.fsl` | `enum Key { A, B }`, no const | an enum map key still works |
| `issue_855_enum_map_key_with_unrelated_const.fsl` | `const Limit = 0` **plus** `enum Key { A, B }` | the diagnostic fires on the *collision*, not on "a const exists in scope" |

Without the second fixture, an implementation that emitted the cause whenever any const was visible
would pass the accepting side. That is the narrowing the pair establishes.

## Test evidence

Reported by the implementer, with the calibration run as a reverted-detector rather than as a reading
of the new tests:

- fix reverted: 1 test selected, **1 failed** on the old message
- fix restored: focused rejecting and accepting controls, 3 passed
- `fsl-core` package tests: passed
- `cargo fmt --check`: passed
- workspace Clippy, `--keep-going` with `-D warnings`: passed
- `./tools/aggregate_changelog.sh check`: passed

**Disclosure about my own verification.** For the other PRs in this batch I re-ran the calibration
myself rather than relying on the reported result. Here I did not: rebasing onto `46ce2db` invalidated
the `fsl-core` build cache and three other cargo tracks were already running on this machine, which is
above the measured concurrency ceiling of two — starting a fourth would have made every timing on the
box untrustworthy, which has already cost this session one unusable measurement. What I did verify
directly is the diff scope (diagnostic and tests only, no resolution change), that the two accepting
fixtures differ in the discriminating way described above, that `loc` is asserted, and that the message
prefix and error kind are unchanged. The calibration itself rests on the implementer's report and on
CI.

Closes #855

### DIFF
--- a/changelog.d/855-const-enum-map-key-shadowing.fixed.md
+++ b/changelog.d/855-const-enum-map-key-shadowing.fixed.md
@@ -1,0 +1,2 @@
+Fixed (#855): map-key diagnostics now explain when a `const` shadows a member of
+the map key enum and advise renaming one of the colliding declarations.
--- a/rust/fsl-core/src/typecheck.rs
+++ b/rust/fsl-core/src/typecheck.rs
@@ -2,7 +2,7 @@
 
 //! Checked-model expression and statement type validation.
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 
 use fsl_syntax::{
@@ -1007,10 +1007,57 @@ fn validate_lvalue(
     }
 }
 
+fn validate_map_key_const_shadowing(
+    target: &LValue,
+    env: &TypeEnv,
+    model: &KernelModel,
+    visible_consts: &BTreeSet<String>,
+    span: Span,
+) -> Result<(), TypecheckError> {
+    match target {
+        LValue::Var(_) => Ok(()),
+        LValue::Index(name, index) => {
+            let (TypeRef::Map(key, _) | TypeRef::Relation(key, _)) = resolve(
+                model,
+                env.get(name)
+                    .ok_or_else(|| error(format!("unknown update target '{name}'")))?,
+            )?
+            else {
+                return Ok(());
+            };
+            let (Expr::Var(member), TypeRef::Named(enum_name)) = (index, key.as_ref()) else {
+                return Ok(());
+            };
+            let Some(TypeDef::Enum { members, .. }) = model.types.get(enum_name) else {
+                return Ok(());
+            };
+            let Some(value) = model.consts.get(member) else {
+                return Ok(());
+            };
+            if !visible_consts.contains(member) || !members.contains(member) {
+                return Ok(());
+            }
+            let (value, actual) = match value {
+                crate::FslValue::Int(value) => (value.to_string(), TypeRef::Int),
+                crate::FslValue::Bool(value) => (value.to_string(), TypeRef::Bool),
+                _ => return Ok(()),
+            };
+            Err(error(format!(
+                "expression of type {actual:?} is not assignable to {key:?}; cause: map key '{member}' resolves to const {member} = {value} ({actual:?}), not to member {member} of enum {enum_name} because a const shadows an enum member of the same name; hint: rename the const or enum member to make the map key unambiguous"
+            ))
+            .with_span(span))
+        }
+        LValue::Field(base, _) => {
+            validate_map_key_const_shadowing(base, env, model, visible_consts, span)
+        }
+    }
+}
+
 fn validate_statement_assignments(
     statement: &Statement,
     env: &TypeEnv,
     model: &KernelModel,
+    visible_consts: &BTreeSet<String>,
 ) -> Result<(), TypecheckError> {
     match statement {
         Statement::Assign {
@@ -1020,6 +1067,7 @@ fn validate_statement_assignments(
         } => {
             let ty = lvalue_type(target, env, model)?;
             ensure_assignable(value, &ty, env, model, *span)?;
+            validate_map_key_const_shadowing(target, env, model, visible_consts, *span)?;
             validate_lvalue(target, env, model, *span)?;
             validate_expression(value, env, model, *span, Some(&ty))
         }
@@ -1030,7 +1078,7 @@ fn validate_statement_assignments(
         } => then_statements
             .iter()
             .chain(else_statements)
-            .try_for_each(|item| validate_statement_assignments(item, env, model)),
+            .try_for_each(|item| validate_statement_assignments(item, env, model, visible_consts)),
         Statement::ForAll {
             binder, statements, ..
         } => {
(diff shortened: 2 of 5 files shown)


rizumita:
## Independent review — no High, no Medium

Reviewed by an agent that did not implement this PR. This was the review I most wanted, because the PR
body states plainly that **I did not re-run its calibration myself** and relied on the implementer's
reported result. Stating a limit is not the same as verifying it. It is now verified.

### The rejecting control is a detector, not a preservation control

With `validate_map_key_const_shadowing` temporarily replaced by a no-op, the test failed (cargo exit
101) with the old message restored:

    produced: invalid init statement: expression of type Int is not assignable to Named("Key") at 12:5
    expected: invalid init statement: expression of type Int is not assignable to Named("Key"); cause: map key 'A' resolves to const A = 0 (Int), not to member A of enum Key because a const shadows an enum member of the same name; hint: rename the const or enum member to make the map key unambiguous at 12:5

Produced beside expected, failing for the intended reason rather than on a build or fixture error.

### The unchanged-contract claim holds, checked against behaviour

CLI output compared before and after the temporary reversion: both give `result: error`, `kind:
semantics`, exit 2, location 12:5, and both preserve the prefix through `Named("Key")`. Only the fixed
behaviour adds the cause and hint. So the claim that kind, exit code, location, and message prefix are
untouched was verified against the prior behaviour, not against the new test's own expectations.

### The accepting pair really narrows

This was the question that decided the review. I asked whether an implementation keyed on *any* visible
const — rather than on the collision — would pass the accepting side. The implementation was temporarily
broadened to reject whenever a visible const exists:

    without_const_exit=0      the no-const fixture still passes
    unrelated_const_exit=101  the `const Limit = 0` fixture fails (fslc status 2)

So the second fixture genuinely rejects a "const merely in scope" implementation. The reviewer also
classified the three correctly rather than letting them all read as detection: the collision fixture is
the detector; the two accepting fixtures are preservation/narrowing controls.

### The wildcard is not a gap

`typecheck.rs:1045`'s `_ => return Ok(())` was my concern — a value type reaching it while carrying the
same shadowing situation would mean the diagnostic silently does not fire. Source const evaluation admits
only `Int` and `Bool`; no other `FslValue` variant can reach this path from accepted FSL. No finding.

### Baseline discipline

Worth recording because it is the trap that produced my worst error today. The reviewer used the **PR head
(`84fd23bd`)** as the restoration baseline, not `origin/main`, and said why: a diff against
`origin/main` shows the intended feature diff and is not a restoration proof. `git diff --exit-code HEAD`
returned 0 after restoring. Earlier today I read a working tree while a mutation was applied, attributed
that state to the baseline, and reported a production defect that did not exist — the same trap, taken
rather than avoided.

PR #870, opened 2026-08-22
test(fslc): cover the refine chain CLI path and its failed_link envelope

### LINKED ISSUE #850: fslc refine の chain 経路（run_refine_chain）に実行可能なテストが一つも無く、failed_link と合成 action_map が無検証
# Problem

`run_refine_chain` (`rust/fslc/src/main.rs:15054`) has **no executable test coverage**, although it
emits a JSON envelope no other code path produces.

# 実測

- 到達経路は一つだけ: `fslc refine IMPL ABS MAPPING <ABS MAP>...` で path operand が3つより多い場合
  (`rust/fslc/src/main.rs:1578-1595`)。operand が3つなら `run_refine`、奇数なら
  `refine chain must list (abs map) pairs after the first mapping` で exit 2。
- リポジトリ内のどのテストも `refine` に4つ以上の path operand を渡していない。
  `rust/fslc/tests/refine_corpus_parity.rs:686` の `run_refine` ヘルパは
  `["refine", implementation, abstraction, mapping, "--depth", N]` に固定。
- 決定的証拠: **`failed_link`** はこの関数だけが出力するキーだが、
  `grep -rn --include='*.rs' 'failed_link' rust/` の結果は `main.rs:15097` の**1件のみ**。
  テストから一度も参照されていない。
- `chain_cli` は無関係。`fslc chain` は `run_project_chain`
  (`rust/fslc/src/main.rs:1539`, 定義は `:3729`) にディスパッチする。

つまりこの関数が出力する以下がすべて無検証である:

- `chain`: 層名の順序列
- `failed_link`: `{from, to, kind}` — どの層間で refinement が破れたか
- 層をまたいで合成される `action_map`(`stutter` の伝播を含む)

# なぜ重要か

多層 refinement の**合成**が `fslc refine` の chain 形式の存在理由そのものである。
`refine` の単一リンク経路は corpus parity で厚くカバーされているのに、
複数リンクを合成する層だけが対照を持たない。

既知の関連挙動として、refine/chain には「層ごとに enum/type 名が衝突するとメンバが落ちる」
という性質があり、合成の正しさは自明ではない。合成 action_map の `stutter` 伝播も
テストが無い。

AGENTS.md:

> A confidently green false negative is more dangerous than a crash.

# 発見経緯

PR #846(clippy 1.98 対応)が `:15066` の `chunks_exact(2)` を `as_chunks::<2>().0` に書き換えた際、
当初「`chain_cli` がこの経路をカバーする」と記述したが、レビューで**偽**と判明した。
書き換え自体は安全だが(奇数 operand は呼び出し前に exit 2 で弾かれるため余りは発生せず、
等価性は N=2/N=5・長さ0-11 の独立プローブで確認済み)、
**無試験の経路を変更していた**という事実が残る。カバレッジ欠落は #846 以前から存在する。

# 期待される挙動

`run_refine_chain` に accepting/rejecting 対照を与える:

1. **accepting**: 3層以上が実際に refines する chain → `result` が成功、`chain` が層名を
   宣言順で列挙し、合成 `action_map` が期待どおり。
2. **rejecting**: 中間リンクが破れる chain → 非ゼロ exit、`failed_link.from` / `.to` が
   **破れた実際のリンク**を指す(最初や最後を機械的に返していないこと)。
   これが無いと「どこで破れたか」の報告が常に正しく見える偽陽性を検出できない。
3. **rejecting**: operand 数が奇数 → exit 2 と既存の診断文。
4. 合成 `action_map` の `stutter` 伝播に対する対照。

`(2)` は必須である。成功経路だけの緑では、この関数がドリフトを検出できることを確立しない。

# 影響範囲

`rust/fslc/src/main.rs:15054` 以降, `rust/fslc/tests/`（新規テストファイル）

rizumita:
## Problem

`run_refine_chain` (`rust/fslc/src/main.rs`) had **no executable test coverage**, although it emits a
JSON envelope no other code path produces.

The route into it is narrow: `fslc refine IMPL ABS MAPPING <ABS MAP>...` with more than three path
operands. Three operands reach `run_refine`; an odd count is rejected with exit 2 by the operand guard
at `main.rs:1578-1585` before the chain function is entered. Nothing in the repository passed four or
more. `fslc chain` is unrelated — it dispatches to `run_project_chain`.

The decisive evidence was that `failed_link` is a JSON key only this function emits, and
`grep -rn --include='*.rs' 'failed_link' rust/` returned exactly one hit: the emit site. So `chain`,
`failed_link`, and the composed `action_map` were all unverified.

## What this PR is not

**No production code changes.** `git diff origin/main -- rust/fslc/src/main.rs` is empty.

I want to be explicit about this because I got it wrong mid-review and briefly directed the work down
the wrong path. I read the working tree with `sed` while a calibration mutation was applied, saw
`"from": abs_name, "to": abs_name`, and concluded `failed_link` reported the same layer for both ends.
`origin/main` in fact reads `"from": impl_name, "to": abs_name`, which is correct. The implementer had
announced that exact mutation beforehand; my "correction" of their label was the error, not their label.
The changelog category was moved from `fixed` to `required` for the same reason — `changelog.d/README.md`
defines `fixed` as "a defect corrected", and no defect was corrected here.

## Coverage added

`rust/fslc/tests/refine_chain_cli.rs`, three tests:

| test | asserts |
|---|---|
| `refine_chain_composes_action_maps_across_three_distinct_layers` | the composed `action_map` across a clean three-layer chain |
| `refine_chain_reports_the_intermediate_link_that_fails` | the exact `failed_link`, including which link failed |
| `refine_chain_rejects_an_unpaired_final_operand` | the CLI's pre-dispatch exit-2 diagnostic |

Fixtures use distinct enum and type names per layer — `ImplPhase`, `MiddlePhase`, `TopPhase`. This
matters: `refine` merges layer types by name, so identical names across layers silently drop members and
would have produced passing tests that proved nothing.

## Mutation-kill evidence

Each mutation was applied to `main.rs`, the failure captured with its actual output, then reverted and
the revert proven with `git diff origin/main -- rust/fslc/src/main.rs` returning empty.

| mutation | observed |
|---|---|
| break the `action_map` composition | actual map `{"advance":"advance_middle","ping":"ping_middle"}` against expected `{"advance":"advance_top","ping":"stutter"}` — failed |
| `failed_link.from` → `abs_name` | actual `{"from":"TopLayer","to":"TopLayer","kind":"abs_state_mismatch"}` against expected `MiddleLayer → TopLayer` — failed |
| bypass the odd-operand guard | exit 0 with a truncated two-layer `refines` envelope — failed |

Each names the wrong value it produced, not merely that a test failed. A test that fails on a mutation
for an unrelated reason is not a calibrated control.

**Scope note on the third row, stated rather than glossed:** that control covers the CLI operand guard
*before* `run_refine_chain`, not malformed-pair behaviour inside it. It is worth having, but it is not
chain-path coverage and should not be read as such.

## Test evidence

- `cargo test --manifest-path rust/Cargo.toml -p fslc-rust --locked` — exit 0, 133 suites, 898 passed,
  0 failed
- all three `refine_chain_cli` tests pass against unmodified `main`, which is the expected result now
  that no defect is claimed
- `cargo fmt --check`, workspace Clippy with `--keep-going` and `-D warnings` — passed
- `./tools/aggregate_changelog.sh check` — PASS

Closes #850

### DIFF
--- a/changelog.d/850-refine-chain-cli-tests.required.md
+++ b/changelog.d/850-refine-chain-cli-tests.required.md
@@ -1,0 +1,2 @@
+Required (#850): Native CLI regression coverage now detects incorrect failed-link
+identities and composed action maps in multi-link `fslc refine` chains.
--- a/rust/fslc/tests/refine_chain_cli.rs
+++ b/rust/fslc/tests/refine_chain_cli.rs
@@ -1,0 +1,225 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Native CLI coverage for `fslc refine`'s multi-link chain route (issue #850).
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use serde_json::{Value, json};
+
+static NEXT_SCRATCH: AtomicUsize = AtomicUsize::new(0);
+
+struct ChainFiles {
+    implementation: PathBuf,
+    middle: PathBuf,
+    top: PathBuf,
+    impl_to_middle: PathBuf,
+    middle_to_top: PathBuf,
+    broken_middle_to_top: PathBuf,
+}
+
+fn scratch(label: &str) -> PathBuf {
+    let ordinal = NEXT_SCRATCH.fetch_add(1, Ordering::Relaxed);
+    let directory = std::env::temp_dir().join(format!(
+        "fslc-issue-850-{label}-{}-{ordinal}",
+        std::process::id()
+    ));
+    std::fs::create_dir_all(&directory).expect("create scratch directory");
+    directory
+}
+
+fn write(path: &Path, source: &str) {
+    std::fs::write(path, source).expect("write fixture");
+}
+
+fn chain_files(directory: &Path) -> ChainFiles {
+    let implementation = directory.join("impl.fsl");
+    let middle = directory.join("middle.fsl");
+    let top = directory.join("top.fsl");
+    let impl_to_middle = directory.join("impl-middle.fsl");
+    let middle_to_top = directory.join("middle-top.fsl");
+    let broken_middle_to_top = directory.join("middle-top-broken.fsl");
+
+    // The three layers deliberately use distinct enum names. `refine` merges
+    // type metadata by name, so reusing a name would let this fixture hide a
+    // layer's members instead of exercising the complete chain.
+    write(
+        &implementation,
+        r"spec ImplLayer {
+  enum ImplPhase { ImplIdle, ImplDone }
+  state { impl_phase: ImplPhase }
+  init { impl_phase = ImplIdle }
+  action advance() { requires impl_phase == ImplIdle impl_phase = ImplDone }
+  action ping() { requires impl_phase == ImplIdle impl_phase = ImplIdle }
+}
+",
+    );
+    write(
+        &middle,
+        r"spec MiddleLayer {
+  enum MiddlePhase { MiddleWaiting, MiddleComplete }
+  state { middle_phase: MiddlePhase }
+  init { middle_phase = MiddleWaiting }
+  action advance_middle() { requires middle_phase == MiddleWaiting middle_phase = MiddleComplete }
+  action ping_middle() { requires middle_phase == MiddleWaiting middle_phase = MiddleWaiting }
+}
+",
+    );
+    write(
+        &top,
+        r"spec TopLayer {
+  enum TopPhase { TopOpen, TopClosed }
+  state { top_phase: TopPhase }
+  init { top_phase = TopOpen }
+  action advance_top() { requires top_phase == TopOpen top_phase = TopClosed }
+}
+",
+    );
+    write(
+        &impl_to_middle,
+        r"refinement ImplToMiddle {
+  impl ImplLayer
+  abs MiddleLayer
+  enum conversion impl_middle ImplPhase -> MiddlePhase {
+    ImplIdle -> MiddleWaiting
+    ImplDone -> MiddleComplete
+  }
+  map middle_phase = convert(impl_middle, impl_phase)
+  action advance() -> advance_middle()
+  action ping() -> ping_middle()
+}
+",
+    );
+    write(
+        &middle_to_top,
+        r"refinement MiddleToTop {
+  impl MiddleLayer
+  abs TopLayer
+  enum conversion middle_top MiddlePhase -> TopPhase {
+    MiddleWaiting -> TopOpen
+    MiddleComplete -> TopClosed
+  }
+  map top_phase = convert(middle_top, middle_phase)
+  action advance_middle() -> advance_top()
+  action ping_middle() -> stutter
+}
+",
+    );
+    write(
+        &broken_middle_to_top,
+        r"refinement BrokenMiddleToTop {
+  impl MiddleLayer
+  abs TopLayer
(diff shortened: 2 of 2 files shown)


rizumita:
## Independent review — one P3, no higher severity

Reviewed by an agent that did not implement this PR. Findings and confirmed non-findings below; every
item marked CONFIRMED was settled by execution, not by reading.

### P3 — scratch fixtures are never cleaned up

`rust/fslc/tests/refine_chain_cli.rs:22`. Each test creates a PID/ordinal directory and never removes
it; 20 retained `fslc-issue-850-*` directories were observed, six fixtures each. **This does not make
verdicts flaky** — names are unique and every fixture is overwritten — so it is untidiness rather than a
correctness problem. Not fixing it in this PR; recording it here so the next person touching the file has
it.

### The retraction in the PR body is confirmed correct

I asked the reviewer to check this independently, because if I had been wrong a *second* time in the
opposite direction, this PR would be missing a fix it should carry. It is not: baseline emits
`"from": impl_name, "to": abs_name`, and `git diff origin/main -- rust/fslc/src/main.rs` returned
empty both before and after calibration.

### Coverage actually reaches `run_refine_chain`

This was the review's most important question, since a test that only exercises the CLI dispatch in front
of the function would make the coverage claim narrower than stated. The first two tests pass **five** path
operands (`IMPL MIDDLE MAP TOP MAP`), so the dispatcher takes `rest.len() == 2` into
`run_refine_chain` at `main.rs:1589`. Neither can be satisfied by `run_refine`'s three-operand route,
nor by `run_project_chain` (reachable only via the `chain` subcommand).

Fixtures use distinct spec, enum, state, conversion, **and enum-member** names across all layers — which
matters because `refine` merges layer types by name, so identical names would silently drop members and
give a passing test that proves nothing.

### The `failed_link` control is a detector, calibrated

Mutation `from: abs_name` made **only** the intended assertion fail, exit 101 at test line 193:

    left:  {"from":"TopLayer","to":"TopLayer","kind":"abs_state_mismatch"}
    right: {"from":"MiddleLayer","to":"TopLayer","kind":"abs_state_mismatch"}

Produced-versus-expected values, not a build, fixture, or unrelated failure — a distinction I asked for
explicitly because conflating them has already cost this session one wrong conclusion.

The reviewer also classified the other two correctly rather than letting them read as coverage: the
action-map test is the acceptance control for chain composition, and the odd-operand test is a
**preservation control** for the pre-dispatch CLI guard, not chain-path coverage, which is what the PR
body says.

### Ambient-state dependence checked

PR #866 turned up a control today that passed on a cold cache and failed on a warm one, because `verify`
memoizes and emits a cache-provenance block. So these tests were run twice in one session: 3 passed each
run, with differing test order. `FSLC_CACHE_DIR` is unset and its implementation lives in verification
code, while these tests invoke `refine`/`run_refine_chain`, so no cache-provenance path is involved.
Scratch reuse cannot affect fixtures because all six files are rewritten.

### Changelog

Category `required` with a matching `Required (#850):` lead word; `aggregate_changelog.sh check`
passed. The tests land in the required lane: this diff evaluates to `run=true,
reason=product-paths-changed`, and the ruleset requires `rust workspace`.

### Stated limits

Only the focused test target was run — no workspace-wide cargo, Clippy, or fmt — because two cargo tracks
is this machine's measured ceiling and both were occupied. CI covers those. `gh pr diff` failed twice on
a GitHub API connection error, so the local diff was cross-checked after confirming the PR head matches
`0c2ed62`.

PR #876, opened 2026-08-22
fix(causal): report a liveness violation found in the initial state instead of a false pass

### LINKED ISSUE #875: causal replay が初期状態の liveness 違反を破棄し expectation を "pass" と報告する(偽陰性)
## Problem

`rust/fslc/src/causal.rs:852` handles only the `Err` arm of `BoundedLivenessMonitor::observe`, so an
`Ok(Some(BoundedLivenessViolation))` at the initial state is discarded. The dropped violation leaves
`verdicts[index]` unset, and result construction at `causal.rs:1162` reports an unset verdict as
**`"pass"`**.

A liveness violation detected at step 0 is therefore reported as a passing expectation.

## Verified directly

```rust
// causal.rs:852 — only Err is handled
if let Err(error) = liveness.observe(&extended, 0) {
```

```rust
// causal.rs:1162 — an unset verdict becomes "pass"
let verdict = if verdicts[index].is_some() { "violated" } else { "pass" };
```

Both read on `db179af`.

## Why the value can be `Some` at step 0

`cargo test -p fsl-runtime zero_window_failure_and_each_static_binding_are_checked -- --exact` passes, and
that test establishes that `within 0` with `before=true, after=false` yields `Some` at step 0
(`rust/fsl-runtime/tests/bounded_liveness.rs:103`, `rust/fsl-runtime/src/lib.rs:1229`). Causal
expectations accept `within 0` and lower it to a kernel `leadsTo`
(`rust/fsl-syntax/src/causal.rs:901`, `rust/fsl-tools/src/causal_expectation.rs:202`), so the reachable
path exists rather than being hypothetical.

With a subsequent event the same drop instead reaches the non-consecutive-step error path, because a
violation returns before advancing `next_step`. So the failure presents differently depending on whether
another event follows — worth knowing when writing the reproduction.

## Severity

This is a false negative in a verification verdict, which `AGENTS.md` names as the most dangerous class:
*"A confidently green false negative is more dangerous than a crash."* Nothing crashes, nothing warns, and
the JSON envelope reports `pass`.

## Why no lint catches it

`observe` returns `Result<Option<BoundedLivenessViolation>, RuntimeError>`. `#[must_use]` on the inner
type does not propagate through `Option`, and `Option` is not itself `#[must_use]` — confirmed with a
`rustc 1.98.0` probe, reproduced independently during this audit. PR #866 annotated the outcome-carrying
types and closed that class for directly-returned values; this shape is the residue, and it is invisible
to the mechanism that fix installs. That gap is #868; this issue is a concrete live instance of it, found
by auditing for it.

Notably `#[must_use]` here would be **ineffective, not merely absent** — so the fix has to be a control or
a caller change, not an attribute.

## Stated limit on the evidence

The runtime-level rejecting case and the source path are confirmed. **The end-to-end causal CLI
reproduction was not completed**: a fresh vendored-Z3 build exceeded eight minutes and was stopped to stay
within this machine's two-cargo-track ceiling. So the external JSON showing `"pass"` for a violating
`within 0` expectation is **source-traced, not CLI-executed**. Completing that is the first task on this
issue, and the finding should not be described as end-to-end confirmed until it is.

## Expected behaviour

`observe`'s `Ok(Some(violation))` at the initial state must set `verdicts[index]` so the expectation
reports `violated`, with the same envelope shape the non-initial path produces.

## Required controls

- **rejecting**: a causal replay whose `within 0` expectation violates at the initial state must report
  `violated`. This control must fail against the current `causal.rs:852` implementation — verify that by
  running it before the fix, not after.
- **accepting**: a clean initial observation still reports `pass`, so the fix does not turn every initial
  state into a violation.
- **envelope parity**: the initial-state violation's envelope agrees with the non-initial path's for the
  same violation. Compare the whole envelope; an exclusion needs a stated reason it cannot be compared,
  not a category label.

## Related

- #868 — the mechanism gap this instance came from.
- The six `rust/fsl-runtime/tests/bounded_liveness.rs` sites (`:71`, `:134`, `:155`, `:157`, `:175`,
  `:177`) call `.expect(...)` and discard the `Option` without asserting `None`. Those are latent
  test-control gaps rather than production failures — they assert less than they appear to — and are
  tracked under #868 rather than here.

rizumita:
## Problem

`rust/fslc/src/causal.rs` handled only the `Err` arm of `BoundedLivenessMonitor::observe` while observing
the initial state, so an `Ok(Some(BoundedLivenessViolation))` at step 0 was discarded. The dropped
violation left `verdicts[index]` unset, and result construction reports an unset verdict as `"pass"`.

**A liveness violation detected at the initial state was reported as a passing expectation.** Nothing
crashed and nothing warned. `AGENTS.md` names this class as more dangerous than a crash.

## Fix

`verdicts` moves above the initial-state loop, and the `if let Err` becomes an exhaustive `match` with no
wildcard:

```rust
match liveness.observe(&extended, 0) {
    Ok(Some(_violation)) => { verdicts[index] = Some(0); }
    Ok(None) => {}
    Err(error) => { return (error_output("internal", ...), 3); }
}
```

This is **the same shape the later-event path already used** — `Ok(Some(_violation)) => verdicts[index] = Some(step)`
at `causal.rs:1145` — rather than a parallel recording path. I checked that specifically, because two
parallel recorders can drift and the envelope-parity control would then be the only thing holding them
together. They are identical, so that control is confirmatory rather than load-bearing.

The exhaustive match matters beyond this fix: a future variant of the return type forces a decision
instead of falling into a wildcard, which is what `AGENTS.md` asks for.

## Why no lint could have caught it

`observe` returns `Result<Option<BoundedLivenessViolation>, RuntimeError>`. `#[must_use]` on the inner
type does not propagate through `Option`, and `Option` is not itself `#[must_use]`. PR #866 annotated the
outcome-carrying types and closed that class for directly-returned values; this shape is the residue and
is invisible to the mechanism that fix installs. **No `#[must_use]` is added here** — it would be
ineffective rather than merely unnecessary, and adding it would read as a fix while changing nothing. The
mechanism gap is #868.

This defect was found by auditing for that gap, which is the argument for having filed #868 as an
abstract mechanism issue rather than closing it as theoretical.

## End-to-end reproduction, which closes the limit #875 was filed with

#875 recorded the external JSON reporting `pass` as **source-traced, not CLI-executed**, and held that
distinction deliberately. It is now executed:

- before the fix, the end-to-end empty-log reproduction **exited 0 and emitted verdict `pass`**; the
  rejecting CLI test failed with `actual String("pass") versus expected violated`, test exit 101
- before the fix, an enabled subsequent event after the same initial violation produced **exit 3**:
  `bounded liveness expected step 0, got 1` — the second failure shape #875 predicted, because a violation
  returns before advancing `next_step`

Both shapes are covered, not just the one that was easiest to reach.

## Controls

Four, in `rust/fslc/tests/causal_cli.rs`:

| control | role |
|---|---|
| `observe_expectations_reports_initial_liveness_violation` | detector |
| `observe_expectations_initial_liveness_violation_skips_later_events` | detector, subsequent-event shape |
| `observe_expectations_initial_and_later_liveness_violations_have_matching_envelopes` | envelope parity |
| `observe_expectations_keeps_clean_initial_liveness_pass` | accepting |

Envelope parity excludes **only** the observed event counts, which measured `0` versus `6` in the two runs
— the values are reported rather than the field being waved off as "run-dependent". Everything else is
compared. That wording is deliberate: PR #866 shipped an exclusion list transcribed from a struct's field
names in which seven of eight entries named fields absent from both envelopes, weakening nothing while
manufacturing the appearance of a considered decision.

## Calibration, re-run by me rather than taken from the report

Reverted the fix to `Ok(Some(_violation)) => {}` and ran the four controls. Independently reproduced by a
reviewer in a separate worktree against this head:

| | |
|---|---|
| unmodified `causal_cli` | exit 0, 45 passed |
| initial-arm mutation | exit 101, 42 passed / 3 failed |
| — two detectors | `left: String("pass"), right: "violated"` |
| — subsequent-event detector | `left: 3, right: 0`, non-consecutive-step error |
| — accepting control | **passed** |
| restored | full diff exit 0; `causal_cli` exit 0, 45 passed |

**The three detectors do not fail identically, and that is the point.** Two fail on the verdict; the
subsequent-event one fails with `exit 3` and `bounded liveness expected step 0, got 1`. Those are the two
failure modes #875 predicted, so reporting them as one shape would hide that the second control exercises a
different path rather than the first one twice. My first draft of this section did exactly that — I saw
three FAILED lines, read one assertion's detail, and generalised it across all three.

The accepting control **kept passing**, which is right: a clean initial observation should report `pass`
with or without the fix. That asymmetry is what distinguishes the three detectors from the preservation
control, and it is observable only by running the mutation rather than by reading the labels.

Also reported by the implementer and not re-run by me: the complete `causal_cli` target passed 45/45,
`cargo fmt --check`, workspace Clippy with `--keep-going` and `-D warnings`, and
`./tools/aggregate_changelog.sh check`.

Closes #875

### DIFF
--- a/changelog.d/875-causal-initial-liveness.fixed.md
+++ b/changelog.d/875-causal-initial-liveness.fixed.md
@@ -1,0 +1,2 @@
+Fixed (#875): causal expectation replay now reports a `within 0` liveness
+violation found in the initial state instead of returning a false `pass`.
--- a/rust/fslc/src/causal.rs
+++ b/rust/fslc/src/causal.rs
@@ -841,6 +841,9 @@ fn run_causal_observe_expectations(
         }
     }
 
+    // Track per-expectation verdicts and violation step.
+    let mut verdicts: Vec<Option<usize>> = vec![None; compiled.len()];
+
     // Observe initial state (step 0).
     let init_state = monitor.state.clone();
     for (index, (expectation, liveness)) in compiled
@@ -849,22 +852,26 @@ fn run_causal_observe_expectations(
         .enumerate()
     {
         let extended = extend_with_ghost(&init_state, expectation, "", &spec_model);
-        if let Err(error) = liveness.observe(&extended, 0) {
-            return (
-                error_output(
-                    "internal",
-                    &format!(
-                        "liveness observe init for '{}': {error}",
-                        compiled[index].id
+        match liveness.observe(&extended, 0) {
+            Ok(Some(_violation)) => {
+                verdicts[index] = Some(0);
+            }
+            Ok(None) => {}
+            Err(error) => {
+                return (
+                    error_output(
+                        "internal",
+                        &format!(
+                            "liveness observe init for '{}': {error}",
+                            compiled[index].id
+                        ),
                     ),
-                ),
-                3,
-            );
+                    3,
+                );
+            }
         }
     }
 
-    // Track per-expectation verdicts and violation step.
-    let mut verdicts: Vec<Option<usize>> = vec![None; compiled.len()];
     let mut events_observed = 0_usize;
     let events_unmapped = 0_usize;
 
--- a/rust/fslc/tests/causal_cli.rs
+++ b/rust/fslc/tests/causal_cli.rs
@@ -730,6 +730,90 @@ fn run_observe(extra: &[&str]) -> (Value, i32) {
     run_cli(&args)
 }
 
+const INITIAL_LIVENESS_EXPECTATION: &str = r#"causal ImmediateLiveness {
+  uses ops from "incident_system.fsl"
+
+  timebase day
+  horizon 1
+
+  clock ops_clock {
+    kernel ops
+    1 tick = 1 day
+  }
+
+  expectation E_Immediate {
+    trigger predicate ops { guardrails >= 1 }
+    response predicate ops { alert_precision >= 4 }
+    within 0
+    clock ops_clock
+  }
+}
+"#;
+
+fn observe_initial_liveness(initial_guardrails: u8, log: &Path) -> (Value, i32) {
+    let scratch = tempfile_dir();
+    let model = scratch.join("immediate_liveness.causal.fsl");
+    let companion = scratch.join("incident_system.fsl");
+    std::fs::write(&model, INITIAL_LIVENESS_EXPECTATION).expect("write causal model");
+    let kernel =
+        std::fs::read_to_string(repository_root().join("examples/causal/incident_system.fsl"))
+            .expect("read companion kernel")
+            .replace(
+                "guardrails = 0",
+                &format!("guardrails = {initial_guardrails}"),
+            );
+    std::fs::write(&companion, kernel).expect("write companion kernel");
+
+    run_cli(&[
+        "causal",
+        "observe-expectations",
+        model.to_str().expect("utf-8 model path"),
+        "--from-log",
+        log.to_str().expect("utf-8 log path"),
+        "--mapping",
+        OBS_MAPPING,
+        "--scope",
+        OBS_SCOPE,
+        "--period-start",
+        "2026-01-01",
+        "--period-end",
+        "2026-03-31",
+    ])
+}
+
+fn empty_observation_log() -> PathBuf {
+    let log = tempfile_dir().join("empty.jsonl");
+    std::fs::write(&log, "").expect("write empty observation log");
+    log
+}
+
+fn later_enabled_observation_log() -> PathBuf {
+    let log = tempfile_dir().join("later.jsonl");
+    std::fs::write(
+        &log,
+        r#"{"action":"tune_alerts","params":{},"state":{"guardrails":1,"alert_precision":4,"mttr_hours":24}}
+"#,
+    )
+    .expect("write later observation log");
+    log
+}
+
+fn remove_observed_event_counts(envelope: &mut Value) {
+    envelope
+        .as_object_mut()
+        .expect("observation envelope")
+        .remove("events_observed");
+    for expectation in envelope["expectations"]
+        .as_array_mut()
+        .expect("expectations")
+    {
+        expectation["event_counts"]
+            .as_object_mut()
+            .expect("event counts")
+            .remove("observed");
+    }
+}
+
 fn generated_evidence_paths(directory: &Path) -> (Vec<PathBuf>, Vec<PathBuf>) {
     let mut evidence = Vec::new();
     let mut lifecycle = Vec::new();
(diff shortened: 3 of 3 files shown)


rizumita:
## Independent review — no soundness finding, one correction to my PR body

Reviewed by an agent that did not implement this. Calibration reproduced in its own worktree against PR
head `fd97f22`, not against `origin/main`.

### Correction to this PR body, which is mine

I wrote the calibration as three detectors failing with `left: String("pass") / right: "violated"`. That
is wrong for one of them. The subsequent-event detector at `causal_cli.rs:927` fails differently:

    left: 3, right: 0    — with the expected non-consecutive-step error, `bounded liveness expected step 0, got 1`

The control is correct; my description conflated the two failure shapes. What happened: my own calibration
run printed three FAILED lines but only the first thread's assertion detail was visible in what I read, and
I generalised that one assertion across all three. Same class of error as the others I have made today —
asserting uniformity I had not checked per case.

The distinction matters because the two shapes are the two failure modes #875 predicted. Reporting them as
one hides that the second control exercises a different path rather than the first one twice.

Corrected calibration, reproduced independently:

| | |
|---|---|
| unmodified `causal_cli` | exit 0, 45 passed |
| initial-arm mutation | exit 101, 42 passed / 3 failed |
| — two detectors | `left: String("pass"), right: "violated"` |
| — subsequent-event detector | `left: 3, right: 0`, non-consecutive-step error |
| — accepting control | **passed** |
| restored against `fd97f227` | full diff exit 0; `causal_cli` exit 0, 45 passed |

### Completeness of the fix — confirmed

This was my main concern: fixing one site while a sibling keeps the same defect leaves the false negative
in place. Checked and clear. `causal.rs:855` is exhaustive recording `Some(0)`; the only sibling causal
call at `:1143` has the same exhaustive shape recording `Some(step)`; the two replay call sites in
`main.rs` handle `Some(violation)` with detailed failure output. **None drops it.**

### The discarded violation detail — pre-existing and consistent

I asked whether storing only the step loses something the envelope should carry. It does, and it does so in
**both** paths: the causal envelope retains only a pass/violated verdict, and no current
`causal-observation.v0` contract field carries violation step, property, bindings, or deadline. That is a
pre-existing diagnostic limitation applied consistently, not something #876 introduces. Named rather than
left implicit, which is what I wanted.

### Envelope parity's exclusion — verified real

Only `events_observed` and per-expectation `event_counts.observed` are removed. The empty log is asserted
as 0; the checked-in later log has six records and direct CLI execution produced `events_observed: 6`,
`event_counts.observed: 6`. Both values are real, not placeholders. After those removals the **complete**
envelopes compare equal. This is not the #866 shape, where seven of eight exclusions named fields absent
from both envelopes.

### The design doc already required this

`docs/DESIGN-causal.md:737` already requires feeding the monitor at each step and reporting pass/violated.
So the fix **restores** an accepted contract rather than changing one — the code was diverging from a
written requirement, which is the strongest possible framing for a soundness fix. `docs/LANGUAGE.md`
documents verify-expectations and states no contradictory initial-state rule. No language syntax,
declaration, LSP, generated snapshot, frozen-Python, or conformance-registration update applies.

### Stated limits

Workspace-wide cargo, fmt, and Clippy were skipped under the two-cargo-track ceiling; CI covers them
(18 SUCCESS / 3 SKIPPED). `gh pr diff` hit a GitHub API connection error, so the exact fetched PR head and
the local `origin/main...HEAD` diff were verified instead.

PR #878, opened 2026-08-22
fix(types): reject Map<Int, V> at check time instead of failing to terminate in verify

### LINKED ISSUE #862: Map<Int, _> のキー領域が spec の全整数定数の最大値から導出されるため、無関係な定数を足すと既存スペックの初期化被覆が壊れる
# Problem

`Map<Int, _>` のキー領域は **spec 自身の整数定数から導出**される。
`KernelModel::map_key_values` が `TypeRef::Int` に対して `0..=max(integer const)` を
**実体化**するため、**用途の無い定数を1つ足すだけで既存スペックの map 初期化が
不完全と判定される**。

# 実測（2026-08-22、PR #860 の2人目独立レビュー中）

`Map<Int, _>` を使うコーパススペックは2件のみ（`specs/cart_fixed.fsl`、`specs/cart_buggy.fsl`）。
どちらも `const MAXU = 1` / `const MAXI = 1` を持ち、`forall i in 0..MAXI` で初期化する。

| 追加した定数 | 結果 |
|---|---|
| `const UNUSED_NEG = -99` | **変化なし**（fixed は proved、buggy は `NoNegativeStock` 違反のまま） |
| `const UNUSED_BIG = 1000000` | **両方が探索前に失敗** — `cart`, `stock` の partial-init エラー |

100万の probe は 1.3 秒で exit 2。領域は線形に伸びる。

負の定数が無害なのは `0..=max` が 0 未満を含まないため。

# なぜ驚くべき性質か

作者から見て、**まったく無関係な定数の宣言が、別の状態変数の初期化被覆を壊します**。

```fsl
const MAXI = 1
const RETRY_LIMIT = 1000000   // ← 別の用途で追加しただけ

state { stock: Map<Int, Int> }
init { forall i in 0..MAXI: { stock[i] = 1 } }   // ← これが不完全になる
```

診断は `init does not assign state variable(s): stock` であり、
**原因となった `RETRY_LIMIT` には一切言及しません**。

# これは #860 の回帰ではない

`rust/fsl-core/src/model.rs` は #860 で無変更であり、`map_key_values` の
`TypeRef::Int` 特別扱いは既存の受理済み挙動である。#860 は
`rust/fsl-runtime/src/explicit.rs` に整数 const キーの解決を足しただけで、
この性質はその前から存在する。

（#860 の作業中、この関数を `domain_values` に差し替えた版が `specs/cart_buggy.fsl` を
壊し、差し替えを revert した経緯がある。つまり `map_key_values` の `Int` 特別扱いは
**コーパスが依存している**。単純に削除はできない。）

# 期待される挙動

方針の決定が先に必要である。

**案 A: 診断を原因に即したものにする。** 部分被覆エラーが `Map<Int, _>` に対して
出るとき、領域がどの定数から導出されたかを述べる。例:
`init does not assign state variable(s): stock (the Map<Int, _> key domain is 0..=1000000, derived from const RETRY_LIMIT; declare a bounded key type instead)`。
最小の変更で、作者が原因に到達できる。

**案 B: `Map<Int, _>` を非推奨にし、有界なキー型を要求する。**
レビュアーは "a known deprecated Map<Int> behavior" と表現した。
実際に非推奨として文書化されているかを確認し、されていれば
`check` 時に警告、最終的に拒否へ進める。コーパス2件の書き換えを伴う。

**案 C: 領域導出を宣言的にする。** 定数の最大値からの暗黙導出をやめ、
`Map<Int, _>` の初期化被覆は明示的な有界型でのみ証明可能とする。
最も筋が通るが、後方非互換であり `map_key_values` に依存するコーパスを壊す。

いずれの案でも:

- **rejecting control**: 無関係な大きな定数の追加が被覆判定を変えることを固定する
  （案 A なら診断文、案 B/C なら拒否）。
- **accepting control**: 負の定数、および有界キー型を使うスペックが影響を受けないこと。

# 未実測

- `Map<Int, _>` が実際に「非推奨」として文書化されているかは未確認。
  レビュアーの表現をそのまま引用しており、`docs/LANGUAGE.md` 等での確認が必要。
- 定数の最大値がさらに大きい場合（1e9 など）にメモリまたは時間で破綻するか。
  100万で 1.3 秒という測定のみ。

# 影響範囲

`rust/fsl-core/src/model.rs`（`map_key_values`）、
`rust/fsl-runtime/src/explicit.rs`（診断文）、`docs/LANGUAGE.md`、
`specs/cart_fixed.fsl`、`specs/cart_buggy.fsl`

rizumita:
## Problem

`Map<Int, V>` was documented as deprecated and warned about at `check` time, but still accepted. Past the
ignored warning, `verify` materialises `0..=max(integer const)` as the key domain and **does not
terminate**.

Measured on an idle machine with the same official Release `fslc 4.3.0` the issue used, with an explicit
timeout distinguishing "returned" from "did not return" by exit code:

| added constant | elapsed | result |
|---|---|---|
| none | 3.7s | `verified` |
| `UNUSED = 100` | 2.5s | `verified` |
| `UNUSED = 1000` | 6.6s | `verified` |
| `UNUSED = 10000` | **>60s** | **did not return** |
| `UNUSED = 1000000` | **>100s** (>900s on a debug build) | **did not return** |

Note on those numbers: `100` at 2.5s beat the 3.7s baseline, so the low end is within noise and no
monotonicity is claimed. The decisive datum is `1000` → `10000`.

**This corrects #862's own premise.** The issue records `1000000` producing a clean exit 2 in 1.3s and
proposes improving the partial-init diagnostic to name the causing constant. Neither holds: no clean exit
2 was reproducible at any magnitude, so that option is inert — the message is never printed. `check` also
returns `result: ok` even at `1000000`, so the issue's "fails before search" describes only `verify`.

Non-termination is also a direct violation of `docs/DESIGN-v1.md`'s G3, *Verification responsiveness —
returns in seconds*.

## Fix

Reject **every grammar-reachable `Map<Int, _>` occurrence** at `check` time — nested map keys and values,
`Set`/`Seq`/`Option` elements, both relation endpoints, and struct fields — with a located diagnostic naming
the replacement. `type T = Map<Int, _>` is **not** an alias bypass: `type` declarations accept ranges, so
that spelling fails to parse at 2:19 with `expected '..'`, measured rather than assumed.

The first revision of this PR rejected only the outermost type, and an independent review found the hole:
`Map<Id, Map<Int, Bool>>` was accepted because `reject_map_int_key` inspected `ty` and its key without
recursing into the value. Since `default_value(Map(key, value))` calls `default_value(value)` per key, the
blowup there is **multiplicative** — `2 × (max_const + 1)` against the rejected top-level case's
`max_const + 1`. The nested form was worse than the form the check rejected, and this PR body previously
claimed a coverage it did not have. The walk is now an exhaustive `match` over `TypeExpr` with no wildcard,
so a future variant forces a decision.

Struct fields turned out to be a third case: grammar-reachable, and previously producing only a generic
non-scalar diagnostic rather than the bounded-key guidance.

The located diagnostic:

```
result: error | kind: type | loc {line: 5, column: 5}
Map<Int, ...> on 'stock' is rejected; use a bounded domain type as key;
declare `type K = 0..<max>` and use `Map<K, ...>`
```

`rust/fsl-core/src/model.rs` gains an independent `reject_map_int_key` plus one call from the state-field
validation site. **`map_key_values` is untouched** — during #860 a revision replacing its `TypeRef::Int`
special case with `domain_values` broke `specs/cart_buggy.fsl` and was reverted, so the corpus depends on
it and this change deliberately does not disturb it.

The guidance rides in `message` because the semantic-error envelope has no `hint` field:
`render_semantic_error` takes `message`, `loc`, `name_resolution` only. Same constraint #869 hit. The
wording keeps the old warning's continuity phrase — *use a bounded domain type as key* — and says
"rejected" rather than "deprecated" because the behaviour is intentionally stronger.

## Corpus preservation, which was the stop condition

`specs/cart_fixed.fsl` and `specs/cart_buggy.fsl` migrate to bounded `UserId`/`ItemId` keys, and
`cart` moves from an `Int` with a `-1` sentinel to `Option<ItemId>`.

That expansion beyond the key type put the counterexample at risk, so I measured **both sides** rather
than reasoning that an unchanged bug implies an unchanged violation:

| | verdict |
|---|---|
| pre-rewrite, Release 4.3.0, original spec | `violated \| NoNegativeStock \| step 4` |
| post-rewrite, this branch | `violated \| NoNegativeStock \| step 4` |
| `cart_fixed.fsl` post-rewrite | `proved` |

**Same violation, same step.** The bounded key type and the `Option` value representation together
preserve the exact counterexample, not merely some violation that also happens to trip the invariant.
That step number is what will tell a future reader they broke something by narrowing `ItemId`.

I had asked for the `Option` change to be reverted as out-of-scope, on the grounds that bundling a
representation change makes the diff hard to review for the one property that matters. Measuring both
sides closed that concern directly, so it stays.

`Map<Int` appears nowhere else under `specs/` or `examples/`, so no other corpus spec is affected.

## Controls, calibrated — and run by the reviewer, not the implementer

Eight rejecting controls, one per grammar position, each asserting the message, `kind: "type"`, the location,
and exit 2 — not merely a non-zero exit. Plus a **nested bounded** accepting control, which matters because a
recursion that is too broad would pass every rejecting control while breaking legitimate specs.

The calibration was executed by the agent that found the nested hole, not by the implementer. That ordering
is deliberate: reviewer-run calibration is stronger evidence than implementer-run.

A correction to how this was counted: I originally asked for "eight one-at-a-time arm removals", one per
position. That number came from counting positions, and the code has **arms**, not positions —
`Set | Seq | Option` share one arm and the relation endpoints are two operands of one expression. Splitting
the shared arm to isolate `Set` from `Seq` would have restructured code that is not structured that way, so
it would have calibrated a hypothetical shape rather than the shipping one. The shared arm was removed once
and **all three** of its controls failed together, which is the true claim.

| removed | result |
|---|---|
| nested map-**key** recursion | only `..._nested_in_map_keys` failed; produced `result:"ok"`, exit 0 |
| nested map-**value** recursion | only `..._nested_in_map_values` failed; produced `result:"ok"`, exit 0 |
| `Set \| Seq \| Option` shared arm | **exactly three** failed — set, seq, option (30 pass / 3 fail) |
| relation-**source** operand | only `..._relation_sources` failed (32 / 1) |
| relation-**target** operand | only `..._relation_targets` failed (32 / 1) |
| struct-field rejection call | only `..._in_struct_fields` failed (32 / 1), producing `struct field 'Payload.inner' has non-scalar type` — the pre-existing generic diagnostic, a **different** produced value from the other five |

Every restore was proven by a plain `git diff --exit-code` returning empty, not by a `grep -c`: a removal
leaves nothing to count, so a zero count would have been vacuous there.

**The accepting control passed under every one of the six mutations**, and that asymmetry is what separates
the detectors from it. Two outcomes would have blocked this PR and neither occurred: no removal left all 33
tests green (which would mean an arm with no control at all), and no removal failed a control other than its
own (which would mean the names and the coverage had drifted apart).

## Additional controls

- **rejecting**: asserts the exact message, `kind: "type"`, the location, and exit 2 — not merely a
  non-zero exit. Calibrated against removal of the validation call: it failed with `actual message null`
  against the expected rejection, while the pre-change CLI exited 0 emitting the old deprecation warning.
- **accepting**: a `type K = 0..N` + `Map<K, _>` fixture stays `ok` with zero warnings.
- `corpus_check_sweep` (required) passes, and
  `explicit_and_bmc_agree_on_every_accepted_top_level_corpus_spec` passes in 139.57s — the two-evaluator
  agreement invariant holds on the rewritten corpus.

### On the message-text classifier

`verification_output.rs:229` selects the JSON error `kind` by message prefix, following the existing arms'
style; it does not affect the exit code. I asked whether anything fails if the diagnostic text and that
predicate drift apart, and then verified the answer rather than accepting it: rewording
`is rejected` → `is not allowed` without touching the predicate makes the control fail with produced
beside expected.

Stated precisely, because the coverage is asymmetric: that catches a **message** change. A change to the
predicate alone is caught only insofar as the test asserts `kind: "type"`.

## Coupled change

`docs/LANGUAGE.md` and `docs/LANGUAGE.ja.md` move from "works but emits a deprecation warning" to
rejected, at corresponding positions; `tools/build_site_reference.py` succeeds, so the 1:1 section
alignment it enforces is intact, and the regenerated `docs/intro/language.{en,ja}.html` are included.
`skills/fsl/reference.md` and `docs/DESIGN-v1.md` updated. The explicit-rendering golden was
**regenerated** from observed output, not hand-edited, and its diff reviewed.

## A real Clippy finding, and where it was

The first push of the nested fix turned `rust checks` red at its "Check Rust workspace (non-test phases)"
step. `rust workspace` also showed red, but its failing step is "Require complete rust-checks and rust-tests
evidence" — a dependent gate reporting the first failure, so that was **one defect showing as two red
squares**.

The cause was that it had not been checked: the implementer verified only `cli_regression` before I had it
commit and stop, so it never reached the workspace gates. The finding itself was **test-only** —
`expected_loc` passed by value where a borrow suffices — fixed by borrowing the same JSON value, with no
assertion weakened. Stating that distinction matters: a formatting problem, a test-only lint, and an unsound
recursion have very different implications, and only the third would have questioned the fix.

## Verification

Measured by the reviewer on `071434a4`:

- `corpus_check_sweep`: exit 0, 3/3, 94.00s
- `explicit_and_bmc_agree_on_every_accepted_top_level_corpus_spec`: exit 0, 1/1, 136.83s
- `cart_buggy.fsl`: exit 1, `violated | NoNegativeStock | step 4`; `cart_fixed.fsl`: exit 0, `proved`
- `map_key_values` SHA-256 identical to `origin/main`: `be93618ef556b27569958bb2fcc56fa5013bbec8525c35664b1e9158127a54f3`
- `cargo fmt --check`, workspace Clippy `--keep-going -D warnings`, `./tools/aggregate_changelog.sh check`: all exit 0
- full `cli_regression`: 33/33

The changelog fragment was rewritten: it described a top-level rejection, and the contract is now every
grammar-reachable position.

Closes #862


rizumita:
## Independent review found a High, and it is against this PR's leading sentence

Reviewed by an agent that did not implement this change. Reported against the pre-rebase `d42b5e7` and
re-anchored to `30659b3`; the executed evidence stays labelled with the SHA it was run on.

**`Map<Int, V>` in a nested position is still accepted.** `reject_map_int_key` matches only the outermost
type, so this exits 0 from both `check` and `verify`:

```
spec NestedMapInt {
  const UNUSED = 10000
  type Id = 0..1
  state { m: Map<Id, Map<Int, Bool>> }
  action stay() { }
}
```

I confirmed the mechanism myself on `30659b3` by reading the three paths rather than accepting the report:

- `rust/fsl-core/src/model.rs:1973` — `matches!(ty, TypeExpr::Map(key, _) if matches!(key.as_ref(), TypeExpr::Int))`
  inspects `ty` and its key. Nothing recurses into the value type, so for `Map<Id, Map<Int, Bool>>` the
  outer key is `Id` and the function returns `Ok`.
- `rust/fsl-core/src/model.rs:~237` — `default_value(TypeRef::Map(key, value))` maps over
  `map_key_values(key)?` and calls `self.default_value(value)?` **for each key**.
- `rust/fsl-core/src/model.rs:193` — `map_key_values` still materialises `0..=max(integer const)` for
  `TypeRef::Int`, untouched here by design.

So the path this PR exists to close is not merely still reachable in the nested position — it is
**multiplicative**. `Map<Id, Map<Int, Bool>>` with `Id = 0..1` materialises `2 × (max_const + 1)` entries
where the rejected top-level case materialised `max_const + 1`. The nested case is worse than the one the
check rejects.

The reviewer measured the probe as exit 0 at `UNUSED = 10000`. That is consistent rather than reassuring:
`2 × 10001` entries with an empty action is cheap for BMC, and this PR's own measurement table puts the
wall between 1000 and 10000 for a spec that does real work.

### Consequence for the contract

The body above says "Reject `Map<Int, _>` at `check` time with a located diagnostic". **That sentence is
currently false.** Either the check becomes total over the type or the sentence narrows to what the check
does. Making it total is the right call: a narrowed sentence leaves the non-termination path reachable,
and G3 in `docs/DESIGN-v1.md` — verification responsiveness — is the property the rejection exists to
protect.

A second possible bypass is being measured rather than assumed: the check runs on the surface `TypeExpr`
before resolution, so `type T = Map<Int, Bool>` followed by `state { m: T }` may reach
`reject_map_int_key` as `TypeExpr::Named("T")` and pass. If so, a purely syntactic recursion does not
close it either. That result will be posted whichever way it comes out.

## What the review confirmed as holding

- **Corpus counterexample preserved, measured on both sides.** `origin/main` and the PR head both exit 1,
  violate at **step 4**, take the identical action/parameter sequence, and end with
  `stock = {"0": 1, "1": -1}`.
- **`map_key_values` is byte-identical to `origin/main`** by SHA-256, and the bounded-map callers stay
  live. That was the #860 constraint this PR deliberately did not disturb.
- **The migrated runtime suite's bounded-domain acceptance test is a detector, not a preservation
  control**: disabling the named-domain arm made it fail with `state variable 'm' assigned more than once
  in init`, and the restore was proven by `git diff HEAD -- rust/fsl-runtime/src/explicit.rs` empty plus
  `grep -c 'if false' … = 0` — a count, not a pattern match.
- **The formatter fixture is a detector**: preserving the colon in the production rewrite failed
  `assertion failed: !once.contains("MAXI:")`.
- Aliases and generics were checked as a class: the language has neither an `Int` type alias nor a generic
  type position, so those were ruled out rather than left unexamined.

The full workspace was not run, by my instruction — two cargo tracks are the measured ceiling on this
machine and the focused commands cover the changed contract. That is a stated limit, not an omission.

**Not merging until the nested case is closed and its controls are calibrated.**

PR #879, opened 2026-08-22
docs(agents): require control labelling, comparison scope, ambient-state, and state-confirmation evidence

rizumita:
## What this changes

Four bullets in `AGENTS.md`'s "Correctness invariants", placed immediately after the existing rule:

> Every formal-to-implementation conformance anchor must include a negative control that rejects a
> known contract-violating trace, transition, or mutation. A green positive path alone does not
> establish that the anchor can detect drift.

They qualify that rule rather than standing alone. Each names a way a control satisfies that sentence
and **still enforces nothing**. No product code, spec, generated artifact, or CI shape changes.

## Why these four, and not a general exhortation

Each was measured in the merge cycle that produced #866, #869, #870, #876, and #878. **None of them is
visible to a green suite** — each needed a mutation, a repeat run, or following a citation to its
source. That is the argument for writing them as obligations rather than advice.

### 1. Detector versus preservation control

#870's odd-operand test covers the CLI operand guard at `main.rs:1578-1585`, which runs *before*
`run_refine_chain`. It establishes that the chain path was not disturbed — not that the chain path is
covered. In #869 the collision fixture is the detector and the no-const/unrelated-const fixtures narrow
it; broadening the implementation to fire on any visible const produced `without_const_exit=0` beside
`unrelated_const_exit=101`, and *that* is what establishes the pair narrows rather than merely passing.

Two supporting clauses come from mistakes made while calibrating, not from theory. A revert verified by
pattern match (`grep 'if in_forall'` in `rust/fsl-runtime/src/explicit.rs`) matched the
`let scope = if in_forall {` line rather than the mutated bare `if in_forall {` guard, while
`grep -c 'if false'` returned 2 — so the revert had not happened. And two simultaneous mutations in one
calibration made the resulting failures unattributable.

The sites are named by content because line numbers cannot carry this: an earlier revision of this body
cited `explicit.rs:688` and `:707`, which match **no committed revision** (`git grep 'in_forall' 72ccb59`
gives 678/690/709/762/770; the reviewer found 687/706 at #869's head and 693/712 at #878's). Those numbers
came from a transient worktree holding both the #862 change and a calibration mutation. That the PR
introducing rule 4 cited line numbers from an unnamed transient state is the best argument for rule 4, so
it is recorded here rather than quietly fixed.

### 2. A comparison control's scope is part of the control

#866's envelope-parity helper excluded eight fields as "run-dependent". **Seven appear in neither
envelope on that path**, so the list excluded nothing while telling the next reader the hole was closed.

An earlier revision of this body, of `e82fa3d`'s message, and of two doc comments merged in #866 said the
list had been transcribed from `BmcOutputOptions`' field names. **That attribution is false** and is
corrected in `5002fcf`, including at its source in
`rust/fslc/tests/issue_843_testgen_walk_violation.rs`. That struct
(`rust/fslc/src/verification_output.rs:78-97` at #866's `0e4adfd`) carries six fields — `depth`,
`deadlock`, `checked_bounds`, `elapsed_s`, `statistics`, `skip_vacuity_probe` — so only three of the seven
are its fields; `engine`, `closure`, `states_explored`, and `max_frontier_width` are not. What is
established is that the list was not read off the emitted JSON. Its origin is not established, so nothing
now claims one.

Earlier in the same PR a hand-picked field list let **two opposite-direction wrong implementations both
pass** — one nulling `loc`/`violating_bindings`/`blame` for an `invariant`, the other nulling
`violating_bindings` for a `type_bound`. Reading cannot distinguish a dead exclusion from a considered
one, which is why the rule requires a check that fails when an excluded key is absent from both sides.

### 3. A control whose verdict depends on ambient state is not a control

The same parity assertion passed on a cold cache and failed on a warm one: `verify` memoizes and emits
a cache-provenance block `testgen` never emits. The first fix pinned the verify-only key set exactly,
which failed on run 1 and passed on run 2 — **the same flake inverted**, because `cache`'s *presence*
varies, not its value. One run cannot distinguish those two states, which is the whole reason the rule
asks for two.

### 4. Confirm what a state *is* before reporting an observation about it

Four reported observations this cycle were wrong about what had been read, not about the code:

| what was read | what it was taken for | consequence |
|---|---|---|
| working tree under an announced calibration mutation | the committed implementation | a production-defect report, retracted in full |
| a worktree ten commits behind `origin/main` | `main` | caught before concluding |
| a Release binary built before the change | the change | `result: ok` from a rejection that does fire |
| a run's creation timestamp | a job's elapsed time | wrong elapsed figures stated twice |

All four are mine, which is why the rule is phrased as an obligation on whoever reports, not on whoever
implements. **Rows two through four are this session's observations and are not recorded anywhere in the
repository** — an independent review searched for durable evidence and found them only here, correctly
reporting them as unconfirmed. The first row is on the record: #870's changelog fragment moved from
`fixed` to `required` when the defect report was retracted.

## Independent review, and what it changed

Reviewed by an agent that did not author this, and its brief disclosed my authorship up front — this PR
has no reviewer who is neither its author nor the orchestrator, and its own subject is controls being
reported as stronger than they are. It returned three Mediums and a Low. All four are accepted; the two
factual ones I confirmed myself by reading rather than taking the report.

| finding | disposition |
|---|---|
| `BmcOutputOptions` provenance is false | **confirmed**, fixed in `5002fcf` including at its merged source |
| `explicit.rs:688`/`:707` match no committed revision | **confirmed**, sites now named by content |
| the `grep -c … == 0` revert rule is not generally satisfiable | **confirmed**, rule rewritten |
| three measurements have no durable record | **accepted**, now marked as session observations |

The third is the one that improved the rule rather than just the prose. A zero count proves a revert only
when the mutation introduced a unique token; a mutation that edits or deletes text leaves a correct revert
with a nonzero count, and a pattern can match a sibling site — exactly the failure the rule's own anecdote
describes. The rule now requires **exact equality to the named baseline**, an empty `git diff`, with the
count as supporting evidence in the unique-token case. #878's independent review had already used both,
and the diff was the load-bearing half; the rule had generalised from the wrong half.

The absolute "one mutation at a time, or a failure is unattributable" is softened for the same reason: a
deliberately compound mutant whose compound expectation is stated **before** it is applied is
attributable. Isolation is the default, not a law.

The review completed all six requested areas read-only, without cargo, and confirmed as clean: #870's
`main.rs:1578-1585` guard does reject odd operands before `run_refine_chain`; #869 does record
`without_const_exit=0` and `unrelated_const_exit=101`; seven of #866's eight exclusions are absent from
both violation envelopes. It also located related-but-not-contradictory guidance in
`.claude/skills/pr-review/SKILL.md:109-112`, `.claude/rules/rust-verifier.md:29`,
`.claude/agents/fsl-soundness-reviewer.md:31`, `CONTRIBUTING.md:81-86`, and
`docs/DESIGN-semantic-mutation-gate.md:55-68`, and found no coupled edit required.

## Verification

`./tools/aggregate_changelog.sh check` PASS and `git diff --check` exit 0. `c1c0edf` was green at 18 pass
/ 3 skip; `5002fcf` re-runs it. The correction touches a Rust test file's doc comments, so the Rust lanes
now cover this PR where they previously skipped it as docs-only — that is a consequence of fixing the
false claim at its source rather than only in prose, and it is the right trade.

The changelog fragment is `required` rather than `documented`: `changelog.d/README.md` defines
`required` as "a check, gate, or piece of evidence made mandatory", and these bullets make specific
evidence mandatory rather than recording a decision.

## What this does not do

It installs no mechanical enforcement. These are contributor and agent rules in the file that already
carries them, enforced by review — the same status as the surrounding bullets in that section. Turning
any of them into a check is a separate decision with its own design note.


rizumita:
## Independent-review record

The PR author was also the orchestrator of the session and disclosed that fact before review. This was an independent model pass within that disclosed session, not an external human or organizational review. I was explicitly asked not to defer to the author's claims and checked the cited repository evidence directly.

### Findings

1. **Medium — the stated `BmcOutputOptions` provenance was false.** The PR body and first commit body said #866's seven dead exclusions were transcribed from `BmcOutputOptions`. The same attribution existed at `rust/fslc/tests/issue_843_testgen_walk_violation.rs:149` and `:191` at `c1c0edf`. Reading `rust/fslc/src/verification_output.rs:78-97` at `0e4adfd` showed six fields; only `depth`, `elapsed_s`, and `statistics` of the seven were fields of that type. `engine`, `closure`, `states_explored`, and `max_frontier_width` were not. Established with `git show 0e4adfd:rust/fslc/src/verification_output.rs` and direct inspection of the test comments. The narrower claim—that seven exclusions appeared in neither compared envelope—was supported. This attribution was corrected in both the PR prose and the merged-source comment.

2. **Medium — the `explicit.rs:688` and `:707` citations did not identify the claimed sites in any cited committed state.** At base `72ccb59`, the two relevant expressions are the bare `if in_forall {` guard at `rust/fsl-runtime/src/explicit.rs:690` and `let scope = if in_forall {` at `:709`; lines 688 and 707 are not those sites. Historical inspection also found lines 687/706 at #869's head and 693/712 at #878's head. Established with `git grep -n 'if in_forall' 72ccb59 -- rust/fsl-runtime/src/explicit.rs` and `git show` of the #869 and #878 revisions. The record now names the sites by content instead of unsupported line numbers.

3. **Medium — the mandatory zero-count revert rule was not implementable universally.** `AGENTS.md:114-120` at `c1c0edf` required proving restoration with `grep -c` returning zero and said multiple mutations made a failure unattributable. A correct revert need not produce zero when the matched text legitimately exists elsewhere, and a predeclared compound mutant can have an attributable compound expectation. This was established by reading the proposed rule against its own two-occurrence `if in_forall` example. The rule now requires exact equality to the named baseline (an empty diff), permits zero-count evidence only for a uniquely introduced token, and permits an explicitly stated compound mutant.

4. **Low — three measurements were unsupported and therefore unconfirmed, not shown false.** The statements that a worktree was ten commits behind, that creation time was mistaken for elapsed time, and that each diagnosis settled in seconds appeared only in the PR/commit prose; there was no repository file-and-line source for them. Searches of the repository and linked PR records did not independently establish them. I reported them as **unconfirmed rather than false**. They are now identified as session observations; #870's changelog category correction is separately on the repository record.

### Scope completed and stated limit

I completed all six requested areas: factual claims and citations; implementability of all four proposed obligations; redundancy and placement within `AGENTS.md`; changelog category/naming/hygiene; coupled guidance surfaces; and wording that overclaimed measurements or generalized beyond its evidence.

I did not execute the cargo-backed historical controls because the author explicitly prohibited cargo while other cargo workloads saturated the machine. This was a stated review limit, not an accidentally omitted check. I verified those historical claims through source, tests, PR records, and recorded command results instead.

### Claims checked and supported

- `rust/fslc/src/main.rs:1578-1585` at #870's head is the CLI operand guard and rejects an odd refine-chain operand list before `run_refine_chain` is called.
- PR #869's record contains the measured pair `without_const_exit=0` and `unrelated_const_exit=101`.
- Seven of #866's eight former exclusions are absent from both compared violation envelopes. The incorrect part was only the asserted `BmcOutputOptions` provenance.
- At `c1c0edf`, `git diff --name-only origin/main...c1c0edf` showed only `AGENTS.md` and the changelog fragment, so the original diff was docs-only.
- `required` matched `changelog.d/README.md` because the change made evidence mandatory rather than merely recording a decision. `./tools/aggregate_changelog.sh check` exited 0.

### Follow-up confirmation at `5002fcf`

- `git diff c1c0edf..5002fcf -- rust/` showed that the only Rust-file change was to `///` documentation comments in `rust/fslc/tests/issue_843_testgen_walk_violation.rs`; no code, assertion, or compared string changed.
- Reading `assert_exclusions_are_live` and `assert_envelope_parity` at `rust/fslc/tests/issue_843_testgen_walk_violation.rs:142-262` confirmed that the corrected comment matches the controls: dead `NOT_COMPARABLE` entries fail; all shared fields except `cost` are compared; testgen-only fields fail; and only named verify-only fields are permitted.

### Related guidance checked

I also checked `.claude/skills/pr-review/SKILL.md:109-112`, `.claude/rules/rust-verifier.md:29`, `.claude/agents/fsl-soundness-reviewer.md:31`, `CONTRIBUTING.md:81-86`, and `docs/DESIGN-semantic-mutation-gate.md:55-68`. They partially overlap the new evidence discipline but do not contradict it and do not require a coupled edit. `CLAUDE.md` incorporates `AGENTS.md` by reference.

With the accepted corrections through `5002fcf`, the two requested follow-up checks were clean.

PR #897, opened 2026-08-25
fix(compat): honor explicit implements overrides in requirements auto action maps

### LINKED ISSUE #786: test: tests/test_dialect_conformance.py has 2 undetected failures (harness cited as a gate is red, unnoticed since CI never runs it)

rizumita:
## Problem

`tests/test_dialect_conformance.py` — the harness AGENTS.md/CONTRIBUTING.md cite for dialect registration — had been silently red on 2 of 217 cases because nothing runs it (#786). Finding 2 (native-only `enum abstraction` corpus exemption) was already resolved on main by `5501a5a` (#837). Finding 1 remained: the frozen Python reference's `_expand_requirements_process` auto-generates an `action_map` entry for every process transition and concatenates it with an explicit `implements { action submit(c, a) -> submit(c) }` override, so `build_refinement` reports `duplicate action map for 'submit'` on `examples/e2e/2_requirements.fsl` — while native `fslc check` agrees with the file's documented `refines` intent.

## Change (frozen Python reference — justified compatibility fix)

This is a genuine evaluator divergence on a corpus file both evaluators must accept (the corpus and native are correct per #494), which is the explicit condition AGENTS.md sets for touching the frozen reference. Two commits:

- `81cb64e`: suppress auto action-map generation for an action name that has an explicit `implements` override.
- `dd1191d` (review round): also skip the auto-map **actor validation** for overridden actions — the independent reviewer reproduced a legitimate override still being rejected by the auto-map's actor check (produced `error/type actor mismatch` vs expected `ok/refines`).

Focused regression tests added in `tests/test_inline_implements_action.py` (13 cases), including an accepting control proving non-overridden actions keep full auto-map generation and actor validation.

## Evidence

- Before: `test_dialect_conformance.py` 1 failed / 222 passed (×2, produced `duplicate action map for 'submit'`); after: **223 passed** (×2). Focused tests 13 passed (×2). Changelog check exit 0.
- Detector mutations (isolated, restores proven by empty diff): 3 by the implementer + reviewer's over-suppression mutation (accepting control fails when suppression over-reaches) + reviewer's override fixture (before: exit 2 `error/type`; after: exit 0 `ok/refines`) and non-override contrast (still exit 2 as expected), each ×2.
- Independent review (separate codex session), two rounds: round 1 found the P2 actor-validation gap, fixed and re-verified; final findings 0 (Approve), full review attached as a comment.

Closes #786

### DIFF
--- a/changelog.d/786-conformance-harness-failures.fixed.md
+++ b/changelog.d/786-conformance-harness-failures.fixed.md
@@ -1,0 +1,4 @@
+Fixed (#786): the frozen Python compatibility reference now lets an explicit
+inline `implements` action correspondence override the same requirements-process
+auto map, so the manual dialect-conformance corpus no longer rejects a valid
+arity-changing refinement mapping.
--- a/src/fslc/dialects.py
+++ b/src/fslc/dialects.py
@@ -924,6 +924,7 @@ def _expand_requirements_with_display(ast, base_dir, bounds_overrides=None):
     out = []
     display_names = {}
     action_maps = []
+    process_action_maps = []
     auto_state_maps = []
     action_aliases = {}
     implements = None
@@ -1006,7 +1007,7 @@ def _expand_requirements_with_display(ast, base_dir, bounds_overrides=None):
             )
             out.extend(expanded)
             auto_state_maps.extend(state_maps)
-            action_maps.extend(maps)
+            process_action_maps.extend(maps)
             for action in maps:
                 action_aliases.setdefault(action[1], []).append(action[1])
         elif tag == "acceptance":
@@ -1043,6 +1044,16 @@ def _expand_requirements_with_display(ast, base_dir, bounds_overrides=None):
         mapping_items.extend(implements["maps"])
         mapping_items.extend(auto_state_maps)
         mapping_items.extend(action_maps)
+        explicit_action_names = {
+            item[1] for item in implements["maps"] if item[0] == "action_map"
+        }
+        # Process transitions synthesize same-name correspondences. An explicit
+        # inline entry is the override for cases such as an arity-changing lower
+        # action; retaining both would turn a valid override into a duplicate.
+        mapping_items.extend(
+            action for action in process_action_maps
+            if action[1] not in explicit_action_names
+        )
         implements["mapping_ast"] = ("refinement", f"{name}Implements{implements['abs']}", mapping_items)
         out.append(("__implements", implements))
     if display_names:
--- a/src/fslc/dialects.py
+++ b/src/fslc/dialects.py
@@ -758,13 +758,15 @@ def _business_action_actor_map(abs_ast):
     return actors
 
 
-def _check_auto_mapped_process_actors(processes, abs_ast):
+def _check_auto_mapped_process_actors(processes, abs_ast, explicit_action_names):
     business_actors = _business_action_actor_map(abs_ast)
     if not business_actors:
         return
     for proc in processes:
         for tr in proc["transitions"]:
             name = tr["name"]
+            if name in explicit_action_names:
+                continue
             biz_actor = business_actors.get(name)
             if biz_actor is None:
                 continue
@@ -1039,14 +1041,18 @@ def _expand_requirements_with_display(ast, base_dir, bounds_overrides=None):
     )
 
     if implements is not None:
-        _check_auto_mapped_process_actors(processes, implements["abs_ast"])
+        explicit_action_names = {
+            item[1] for item in implements["maps"] if item[0] == "action_map"
+        }
+        _check_auto_mapped_process_actors(
+            processes,
+            implements["abs_ast"],
+            explicit_action_names,
+        )
         mapping_items = [("impl", name), ("abs", implements["abs"])]
         mapping_items.extend(implements["maps"])
         mapping_items.extend(auto_state_maps)
         mapping_items.extend(action_maps)
-        explicit_action_names = {
-            item[1] for item in implements["maps"] if item[0] == "action_map"
-        }
         # Process transitions synthesize same-name correspondences. An explicit
         # inline entry is the override for cases such as an arity-changing lower
         # action; retaining both would turn a valid override into a duplicate.
--- a/tests/test_inline_implements_action.py
+++ b/tests/test_inline_implements_action.py
@@ -210,6 +210,54 @@ def test_inline_action_map_overrides_process_auto_map_only():
     ]
 
 
+BUSINESS_PROCESS_ACTOR_OVERRIDE = '''business ActorOverrideBiz {
+  actor Employee, Manager
+  entity Claim
+
+  process Claim {
+    stages Draft, Submitted
+    initial Draft
+    transition submit Draft -> Submitted by Employee
+    transition review Draft -> Submitted by Manager
+  }
+}
+verify {
+  instances Claim = 1
+}
+'''
+
+REQUIREMENTS_PROCESS_ACTOR_OVERRIDE = '''requirements ActorOverrideReq {
+  implements ActorOverrideBiz from "biz.fsl" {
+    action submit(c) -> review(c)
+  }
+
+  process Claim {
+    stages Draft, Submitted
+    initial Draft
+    transition submit Draft -> Submitted by Manager
+  }
+}
+verify {
+  instances Claim = 1
+}
+'''
+
+
+def test_inline_action_map_override_skips_auto_map_actor_validation(tmp_path):
+    _write(tmp_path, {
+        "biz.fsl": BUSINESS_PROCESS_ACTOR_OVERRIDE,
+        "req.fsl": REQUIREMENTS_PROCESS_ACTOR_OVERRIDE,
+    })
+
+    checked = run_check(str(tmp_path / "req.fsl"))
+
+    assert checked["result"] == "ok", checked
+    assert checked["implements"] == {
+        "abs": "ActorOverrideBiz",
+        "result": "refines",
+    }
+
+
 ABS_BRANCH = '''spec AbsBranch73 {
   type CaseId = 0..1
   state { done: Map<CaseId, Bool> }
--- a/tests/test_inline_implements_action.py
+++ b/tests/test_inline_implements_action.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2026 Ryoichi Izumita
 
-"""Inline `implements { }` action correspondence (#73).
+"""Inline `implements { }` action correspondence (#73, #786).
 
 Confirms `refinement_action` can appear inside the requirements-dialect
 inline `implements { }` block (grammar.py `?implements_item`), including
@@ -11,7 +11,10 @@ refinement file + `fslc refine`. The inline desugar (dialects.py
 same `("refinement", ..., mapping_items)` AST that the separate-file path
 parses, so `refine.py` needs no changes.
 """
+from pathlib import Path
+
 from fslc.cli import run_check, run_refine, run_verify
+from fslc.dialects import _parse_file
 
 
 def _write(tmp_path, files):
@@ -178,6 +181,33 @@ def test_inline_action_map_conflicts_with_requirement_maps_clause_is_an_error(tm
     assert checked["result"] == "error"
     assert checked["kind"] == "type"
     assert "duplicate action map for 'pay'" in checked["message"]
+
+
+def test_inline_action_map_overrides_process_auto_map_only():
+    root = Path(__file__).resolve().parents[1]
+    requirements = root / "examples/e2e/2_requirements.fsl"
+
+    checked = run_check(str(requirements))
+    assert checked["result"] == "ok", checked
+    assert checked["implements"] == {"abs": "ExpenseToBe", "result": "refines"}
+
+    ast, _display_names = _parse_file(requirements)
+    implements = next(item[1] for item in ast[2] if item[0] == "__implements")
+    action_maps = [
+        item for item in implements["mapping_ast"][2] if item[0] == "action_map"
+    ]
+
+    assert [(item[1], item[2], item[3]) for item in action_maps] == [
+        (
+            "submit",
+            [("refinement_param", "c", None), ("refinement_param", "a", None)],
+            ("action", "submit", [("var", "c")]),
+        ),
+        ("auto_approve", ["c"], ("action", "auto_approve", [("var", "c")])),
+        ("mgr_approve", ["c"], ("action", "mgr_approve", [("var", "c")])),
+        ("reject", ["c"], ("action", "reject", [("var", "c")])),
+        ("pay", ["c"], ("action", "pay", [("var", "c")])),
+    ]
 
 
 ABS_BRANCH = '''spec AbsBranch73 {


rizumita:
# Issue #786 Finding 1 — independent review of `81cb64e`

## Verdict

**Findings: 1. Request changes.**

The mapping-assembly change is narrow and its same-name suppression behaves as
intended for the reported corpus case. It also retains auto correspondences for
unrelated process actions. However, the frozen Python path still performs the
*auto-mapped* actor check for a process action whose auto correspondence was
suppressed by an explicit inline override. This leaves a valid class of explicit
overrides rejected and makes the changelog claim broader than the implemented
behavior.

Review target: `81cb64ef525d2e318da1a15537a88467fa5192e2`
(parent `f1665c2ebc0b5c68a6a3519de41b89b03727b44d`). Finding 2 is outside this
diff and commit `5501a5a` is an ancestor of the review target.

## Finding 1 — P2: an explicit process-action override still trips the auto actor check

### Claim

An explicit inline correspondence suppresses the generated process
`action_map`, but it does not suppress the separate actor-compatibility check for
that generated correspondence. Consequently, the frozen Python reference can
reject a valid explicit mapping as if the process action were still auto-mapped.
The emitted diagnostic even recommends writing the explicit map that is already
present.

### Citation

- `src/fslc/dialects.py:761-778` checks every requirements-process transition
  against a same-named business action and has no knowledge of explicit action
  names.
- `src/fslc/dialects.py:1041-1056` invokes that check before computing
  `explicit_action_names`; the new filter affects only `mapping_items`.
- `docs/LANGUAGE.md:1377-1387` says explicit entries override auto defaults.
  `docs/LANGUAGE.md:1394-1399` scopes actor compatibility to **auto-mapped**
  correspondences and again says explicit entries take precedence.
- The unchanged native lineage already implements that distinction:
  `rust/fsl-core/src/refinement.rs:897-900` skips auto synthesis when an explicit
  correspondence exists, and `rust/fsl-core/src/refinement.rs:1163-1169` checks
  actors only when `origin == CorrespondenceOrigin::Auto`.
- `changelog.d/786-conformance-harness-failures.fixed.md:1-4` currently states
  without qualification that an explicit inline action overrides the same
  requirements-process auto map.

### Produced

A minimal `/private/tmp` fixture declares business actions `submit` by Employee
and `review` by Manager. Its requirements process declares `submit` by Manager
and explicitly maps `action submit(c) -> review(c)`.

The frozen Python CLI was run twice in the fresh review venv. Both runs produced
exit `2` and:

```text
result: error
kind: type
message: transition 'submit' is auto-mapped to business action 'submit' but its
         actor 'by Manager' does not match the business actor 'by Employee';
         rename the action or write an explicit map
loc: 8:5
```

### Expected

Exit `0`, top-level `result: ok`, and `implements.result: refines`. The authored
correspondence is explicit, maps to the Manager-owned `review` action, and thus
must not be validated as the suppressed same-name auto correspondence.

### Suggested correction

Exclude explicit inline action names from the actor check using the same
provenance boundary as the new process-map filter, and add a focused control in
which a differently named explicit target intentionally resolves a same-name
actor mismatch. Keep the existing auto-only mismatch test as the rejecting
control.

## Checks performed

### Scope and minimality

- `git diff 81cb64e^ 81cb64e`: only
  `src/fslc/dialects.py`, `tests/test_inline_implements_action.py`, and
  `changelog.d/786-conformance-harness-failures.fixed.md`; `+47/-2`.
- No `.fsl` corpus file, native Rust implementation, frozen grammar, registry,
  or Finding 2 implementation changed.
- The new list separation is otherwise appropriately narrow:
  requirement-action `maps` and generated time mappings remain in
  `action_maps`; only process-generated mappings enter `process_action_maps`.
  Duplicate explicit entries and explicit-vs-requirement-action `maps` therefore
  still reach `src/fslc/refine.py:603-606`.

### Reproduction in a fresh temporary venv

The review venv used Python 3.14.7 and a fresh `.[dev]` install.

| Check | Produced | Expected | Exit |
|---|---|---|---:|
| Parent `f1665c2`: `python -m pytest tests/test_dialect_conformance.py -q` | `1 failed, 222 passed`; `error/type`, `duplicate action map for 'submit'`, loc `38:5` | detector rejects the reverted implementation for that duplicate | 1 |
| Target `81cb64e`: same command | `223 passed in 127.92s` | all cases pass | 0 |
| `test_inline_implements_action.py test_req_process.py -q`, run 1 | `12 passed in 7.01s` | pass | 0 |
| Same focused suite, run 2 | `12 passed in 6.28s` | pass | 0 |

The parent and target full-harness runs used separate detached worktrees and the
same venv, with `PYTHONPATH=src` resolving to each named revision.

### Accepting controls and mutation calibration

- `tests/test_inline_implements_action.py:186-210` asserts the exact lowered
  action-map list: explicit `submit` occurs once while `auto_approve`,
  `mgr_approve`, `reject`, and `pay` retain their generated maps. This establishes
  that an override for one name does not suppress different names.
- `tests/test_req_process.py:72-82` uses an empty inline `implements` block and
  establishes that process auto maps remain when no explicit override exists.
  It passed in both focused-suite runs.
- Revert detector: the isolated parent revision reproduced `error/type:
  duplicate action map for 'submit'` where the fixed expectation is a non-error
  agreement result. This is the expected failure, not merely an arbitrary test
  failure.
- New over-suppression mutant: changed the process-map inclusion predicate to
  constant false in an isolated target worktree. The no-explicit-override control
  exited `1`, producing `error/type: missing action correspondence for impl
  action 'submit'` where `result: ok` and `implements.result: refines` were
  expected. The mutant was restored; `git diff --exit-code` returned `0` in both
  the mutant worktree and the primary review worktree.

### Changelog and hygiene

- `.fixed.md` is the correct defect-fix category and its filename/body satisfy
  `changelog.d/README.md` mechanically. Its unconditional wording should be
  revisited with Finding 1.
- `./tools/aggregate_changelog.sh check`: PASS twice, exit `0` both times.
- `git diff --check`: exit `0`.
- Final primary tracked diff before writing this report: empty. The two supplied
  task briefs remained the only pre-existing untracked files.

No cargo command was run, no source/corpus file was retained as modified, and
nothing was pushed or committed.

## Final re-review addendum — `dd1191d`

**Final verdict: 0 remaining findings. Approve.** This addendum supersedes the
request-changes verdict above while preserving the original P2 report and its
reproduction evidence as review history.

Review target:
`dd1191d732d94763dfdd5bf2a425018ee12f6e3d`
(`fix(compat): skip auto-map actor validation for overridden actions`).

### P2 resolution

The fix closes the reported provenance gap at the same boundary used for map
suppression:

- `src/fslc/dialects.py:1043-1051` derives the inline
  `explicit_action_names` set before actor validation and supplies that exact set
  to `_check_auto_mapped_process_actors`.
- `src/fslc/dialects.py:761-780` skips only a process transition whose action
  name is in that set. All other same-name process transitions retain the
  existing business/requirements actor comparison.
- `src/fslc/dialects.py:1059-1062` uses the same set for generated process-map
  suppression, so actor validation and auto-map assembly share one provenance
  boundary.
- `tests/test_inline_implements_action.py:213-258` adds the original reviewer
  fixture as a repository regression control. The pre-existing non-override
  rejection remains pinned at `tests/test_req_process.py:100-112`.

### Produced / expected

The original external `/private/tmp` actor-override fixture was re-used rather
than relying only on the new in-tree test:

| Control | Produced | Expected | Exit |
|---|---|---|---:|
| Explicit `action submit(c) -> review(c)`, run 1 | `result: ok`, `implements: {abs: ActorOverrideBiz, result: refines}`, no warnings | explicit override is accepted as `ok/refines` | 0 |
| Same explicit fixture, run 2 | identical `ok/refines` result | identical pass | 0 |
| Same actor mismatch with no explicit override, run 1 | `error/type`; `submit` auto-map actor Manager != Employee, loc `7:5` | actor mismatch is rejected | 2 |
| Same no-override control, run 2 | identical `error/type` and location | identical rejection | 2 |

The repository's no-override negative test was also run directly twice. Both
runs exited `0` with `1 passed` (0.32s and 0.34s); its assertions require
`result == "error"`, `kind == "type"`, and both sides of the exact actor-mismatch
message. This independently confirms that the skip did not spread to ordinary
auto correspondences.

### Final verification on `dd1191d`

- `PYTHONPATH=src <review-venv>/bin/python -m pytest
  tests/test_dialect_conformance.py -q`: exit `0`, `223 passed in 122.47s`.
- `PYTHONPATH=src <review-venv>/bin/python -m pytest
  tests/test_inline_implements_action.py tests/test_req_process.py -q`, run 1:
  exit `0`, `13 passed in 6.62s`.
- Same focused suite, run 2: exit `0`, `13 passed in 6.49s`.
- `git diff --check` and `git diff --check 81cb64e dd1191d`: exit `0`.
- The fix commit changes only `src/fslc/dialects.py` and
  `tests/test_inline_implements_action.py` relative to `81cb64e`; no corpus,
  native Rust, frozen grammar, registry, or changelog surface changed.

### Checks tried with zero findings

1. Re-ran the exact external fixture that originally demonstrated P2, twice.
2. Ran a matched external no-override actor-mismatch control, twice.
3. Ran the existing exact no-override actor-rejection pytest, twice.
4. Inspected the skip set and verified the same action-name provenance boundary
   controls both actor validation and generated process-map suppression.
5. Ran the full manual dialect-conformance harness (`223 passed`).
6. Ran the focused inline/process sibling suite twice (`13 passed` each).
7. Checked fix scope, whitespace, and retained duplicate/exact-map controls in
   the focused suite.

No product source was changed by this final review, and nothing was pushed or
committed.

PR #919, opened 2026-08-27
revert(production): restore the v4.4.1 candidate tree

rizumita:
## Problem

Promotion PR #917 used `main` as its head, so it carried whatever `main` pointed at when it merged — not the recorded candidate. #916 (`docs(design): accept nested Option full support`) landed on `main` between the candidate SHA and the promotion merge, so `production` now contains it.

`docs/RELEASE.md` §2 requires the promoted `production` tree to match the approved `main` tree. It does not:

- candidate `1552033` tree: `14345fc5f5e60c618b8ec3a265e8c30ce02b8d54`
- `production` `1926343` tree: `943989c3712d85238036853c5bee3ea03ae7497e`

The difference is three docs files only (`docs/DESIGN-nested-option-support.md` +597, `docs/README.md` +1 index line, one changelog fragment) — **no product code, tests, or workflows**. But a changelog fragment on `production` that is absent from the 4.4.1 section is exactly the drift the tree-identity rule exists to catch, so it is corrected rather than waived.

## Change

Reverts `c212de1b` on `production`. Verified: the resulting tree hash is `14345fc5f5e60c618b8ec3a265e8c30ce02b8d54` — **identical to the candidate tree**.

Chosen over a force-push rollback because `production` is protected by the `production promotion gate` ruleset (`non_fast_forward`, `pull_request`, `required_status_checks`, no bypass actors); reverting keeps the protection and the history intact.

## Follow-up (must not be lost)

`#916`'s content is still on `main` and is wanted there. Because this revert lives on `production`, the next promotion will hit the revert-of-revert problem: git will treat the change as already applied and silently keep it deleted. The next promotion must therefore re-apply it explicitly (`git revert` of this commit, or a fresh cherry-pick) and verify `docs/DESIGN-nested-option-support.md` exists on `production` afterwards. Tracked as a follow-up issue.

Refs #917, #916


rizumita:
Superseded by a policy-compliant branch. `production-policy.yml:25` accepts only `main`, `release/vX.Y`, or `hotfix/vX.Y.Z` as promotion heads, so `revert/916-on-production` was rejected by the required `production source policy` check. Reopening the same commit from `hotfix/v4.4.1` (identical tree `14345fc`).

PR #955, opened 2026-08-30
feat(verify): add an explicit-engine action execution profile

### LINKED ISSUE #316: feat(verify): explicit エンジンにアクション実行プロファイル(enabled/fired/no-op 統計)を追加する
# 問題

既存の `action_coverage` は「深さ K 以内に各アクションが一度でも enabled になったか」の bool のみで、**死んではいないが痩せている**仕様部分が見えない:

- enabled にはなるが発火しても状態を変えないアクション(no-op fire)
- ごく限られた状態でしか enabled にならないアクション

TLA+ TLC の coverage statistics(アクション発火回数の統計)に相当するプロファイルがない。issue #212 / #226 で solver-free explicit エンジンが入ったため、探索ループでのカウント収集はほぼ無償で可能になっている。

# 提案

`--engine explicit`(および `--engine auto` が explicit で決定した場合)の実行で、アクションごとに以下を集計し、新フィールド(例: `action_profile`)としてエンベロープに追加する:

- enabled 回数(enabled になったユニーク状態数)
- fired 回数(探索済み状態グラフ上のユニーク (state, action) 遷移数)
- no-op fire 回数(fired のうち `child.state == *state` だったもの)

# 実装方針(2026-07-16 のコードベース調査に基づく)

- 挿入点は `rust/fsl-runtime/src/explicit.rs::verify_explicit_selected` の2ループ: `enabled_by_state` 構築ループ(既存の `action_coverage.insert` の隣)と `next` 構築ループ(`step_selected` 直後)。`ExplicitResult` に `BTreeMap<String, ActionStats>` を追加。
- **決定性**: カウントは「探索済み状態グラフ上のユニーク (state, action) 遷移」で定義する。pop ごとの素朴なインクリメントは `seen` による重複除去のタイミングと整合せず、二重カウント/非決定を生む。探索は `BTreeMap`/`BTreeSet` キー付けで決定的なので、この定義なら native/Worker で一致する。
- **`cost` ブロックに相乗りしない**: `VerificationStatistics` / `SolverMetrics` は `rust/fsl-solver` に属し、`fsl-runtime` の solver 非依存境界(AGENTS.md)に抵触する。独立型を `fsl-runtime` 側に置き、JSON では `cost` と並置する(explicit 実行で `cost` がゼロ埋めされる既存契約は変えない)。
- **レンダリングの2重実装を両方更新**: `rust/fslc/src/verification.rs::add_common_verification`(CLI 専用)と `rust/fslc/src/verification_output.rs`(`fsl-wasm` から共有)は同型ロジックが重複しており、片方だけ変更すると native CLI と Worker で出力が食い違う。
- 契約テスト: `rust/fslc/tests/explicit_engine.rs`、`rust/fslc/tests/issue_226_auto_engine.rs`。下流では `rust/fsl-tools/src/ledger.rs` が `action_coverage` の形を消費しているため構造互換を確認する。
- BMC/induction(symbolic)には適用しない。explicit 専用の診断であることを JSON 上も文書上も明示する。

# スコープ外(明示)

**guard の disjunct 単位の分岐追跡**(バウンド内で一度も真にならない `||` 分岐の検出)は含めない。concrete evaluator の `eval()` は式の途中結果を外部に出す経路を持たず(`rust/fsl-runtime/src/lib.rs::evaluate_action_guards` は単一 Bool)、可視化チャネルの追加は `Monitor::enabled` / `attempt` / `step_selected` / refinement 系の多数の呼び出し元に波及する大規模変更になる。やるなら「`requires` 直下の最上位 `||` のみ・ネストなし」に限定した縮退案を別 issue で設計する。

# 関連する既知ギャップ(本 issue のスコープ外)

`action_coverage.<name>.blocking_requires` は Rust 側で空配列スタブ(`verification.rs` / `verification_output.rs`)であり、Python 参照実装 `src/fslc/bmc.py` の `_blocking_requires_hint` / `_minimize_requires_core` 相当が未移植。プロファイルの隣接領域として認識のみしておく。

# 結合変更

`docs/LANGUAGE.md`、`skills/fsl/reference.md`、design note、`CHANGELOG.md`。

# 先行例

TLA+ TLC の coverage statistics(アクション発火回数・未実行 disjunct の統計)。

rizumita:
Resolves #316.

## Problem

`action_coverage` only records whether each action became enabled at least once
within depth K. That hides the parts of a spec that are **thin rather than dead**:
an action that becomes enabled and fires without changing state, or one that is
enabled only in a narrow corner of the state space.

## Contract change

`--engine explicit` (and `--engine auto` when it resolves to explicit) now emits
`action_profile`, one entry per action:

- **enabled** — unique states in which the action was enabled
- **fired** — unique `(state, action)` transitions in the explored state graph
- **no_op** — fired edges whose successor state equals the source

Counts are defined over the explored state graph, not incremented per pop, so they
do not depend on when `seen` deduplicates. `action_coverage` keeps its existing
shape — `rust/fsl-tools/src/ledger.rs` consumes it — and `action_profile` is added
alongside rather than replacing it. Symbolic BMC and induction are unaffected; the
profile is explicit-only, and that is stated in the JSON and the docs.

`action_profile` does not ride in the `cost` block: `VerificationStatistics` and
`SolverMetrics` belong to `rust/fsl-solver`, and `fsl-runtime` must stay
independent of the solver.

## The issue's premise about a dual renderer no longer holds

#316 asked for both rendering implementations to be updated, warning that changing
one would make native CLI and Worker output diverge. **That duplication is already
gone** and was verified before relying on it:

- `add_common_verification` does not exist
- `rust/fslc/src/verification.rs` *calls* `fslc_rust::verification_output::render_bmc_output`
  — one implementation, shared, not two with the same shape
- `fsl-wasm` has no explicit-engine call site

So the control the issue implied — mutate one side, watch the comparison fail — has
no premise left. The shared renderer's own isolation test covers that boundary
instead. Recorded on the issue rather than silently reinterpreted; the issue's other
directions (insertion points, the determinism definition, the `cost` boundary,
`action_coverage` compatibility) all still hold.

## Calibration

The control is `explicit_action_profile_is_deterministic_and_keeps_coverage_compatible`
in `rust/fslc/tests/explicit_engine.rs`, which compares the whole profile against a
declared expectation **and** compares two runs of the same spec against each other.

The fixture `explicit_action_profile_is_deterministic...` is built so the three
counters do not move together — otherwise breaking one of them would leave no trace:

| action | enabled | fired | no_op |
| --- | --- | --- | --- |
| `advance` | 2 | 1 | 0 |
| `limited_reset` | 1 | 0 | 0 |
| `set_one` | 1 | 1 | 0 |
| `stutter` | 2 | 2 | 2 |

`advance` pins `enabled != fired` (enabled at depth 1, but its out-edges are not
explored there). `stutter` pins `no_op != 0` (two parameter instances, both
self-loops). `fired` counts before child-state dedup, so `set_one` converging on
`advance`'s successor still registers its own firing.

Each counter was then broken **alone**, reverting to SHA-256 `6a0b6eca44f054a0`
between mutations:

| mutation | expected | produced |
| --- | --- | --- |
| `stats.no_op += 1` → `+= 0` | test passes | `exit=101` — `stutter.no_op` 2 → **0**; every other action, and `stutter`'s own `enabled`/`fired`, unchanged |
| `stats.fired += 1` → `+= 0` | test passes | `exit=101` — `advance` 1→**0**, `set_one` 1→**0**, `stutter` 2→**0**; `limited_reset` stays 0 because it never fires; `enabled`/`no_op` unchanged |
| `.enabled += 1` → `+= 0` | test passes | `exit=101` — all four actions' `enabled` → **0**; `fired`/`no_op` unchanged |

Each mutation moves exactly the counter it targets and leaves the other two intact,
which is what makes this one test a detector for three separate properties rather
than for their sum.

## clippy caught what the tests did not

After rebasing, the gates were read from their logs rather than inferred:

```
fmt          exit=0
clippy       exit=101   docs for function which may panic missing `# Panics` section
                        --> fsl-runtime/src/explicit.rs:128  verify_explicit_selected
fsl-runtime  exit=0     (74 tests, all passing)
```

The implementation added two `.expect(...)` calls, which made
`verify_explicit_selected` panic-capable. **The `.expect`s were not removed to
silence the lint** — the panic condition is documented instead: it would mean the
Monitor and the model disagreed about which actions exist, which no input can reach
through the public API and which the profile counters could not be attributed under.
That is worth stating where a future reader will need it.

74 tests passed while clippy failed. Running focused tests and skipping workspace
clippy would have sent this to CI.

After the fix: `fmt exit=0`, `clippy exit=0`.

## Verification

| gate | exit |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | 0 |
| `cargo test -p fsl-runtime --locked` | 0 (74 tests) |
| `cargo test -p fslc-rust --locked --test explicit_engine` | 0 (17 tests) |

## Not in scope

Per-disjunct guard tracking (detecting `||` branches never true within the bound)
is excluded, as #316 states: the concrete evaluator's `eval()` has no channel for
intermediate results, and adding one reaches many call sites. The known
`action_coverage.<name>.blocking_requires` empty-array stub is likewise untouched.

Authorship note for review: this branch was produced under my orchestration and the
calibration above was executed by me, not by an independent reviewer — please read
it as an authored change.


rizumita:
Follow-up correction and integration fix for #316:

- `issue_663_kernel_projection_owner::every_key_a_curated_verify_corpus_emits_is_classified` failed on this branch because `verify` emitted the new `action_profile` key without a registered projection fate. The branch produced 5 passed / 1 failed, while the `78d4ae6` main baseline produced 6 passed / 0 failed.
- I had previously described that full-suite failure as pre-existing without first reproducing the same test on main. That classification was wrong; the failure was introduced by #316. A claim that a failure is pre-existing needs a same-test baseline measurement.
- `action_profile` is now classified as `Dropped` in `fslc_rust::outcome::classify_kernel_key`. `action_coverage` remains `Projected` because it qualifies confidence in the verdict (for example, whether a named action was ever enabled). In contrast, `action_profile` contains engine-specific counts from the concrete explicit-state exploration. Those counts are useful CLI run diagnostics, but they are neither replayable verdict evidence nor part of the Public Kernel model contract, so projecting them into `db`/`domain` kernels would make that contract depend on the chosen verification engine.
- The subsequent full `fslc-rust` run exposed a second #316 integration omission: the shared explicit-renderer normalized golden did not yet include `action_profile`. I regenerated it through the test-owned `FSLC_UPDATE_VERIFICATION_GOLDEN` path and verified that deleting `action_profile` from old and new JSON makes the documents exactly equal.

Verification for commit `ed47db6`:

- focused issue #663 suite: `exit=0` (6 passed)
- full `fslc-rust` suite: `exit=0`
- workspace/all-targets Clippy with `-D warnings`: `exit=0`

Issue #984, opened 2026-09-04
perf(ci): evaluate trimming debug info to shrink the semantic-mutation / rust-native-z3 Actions caches


rizumita:
## Problem

The two largest Actions cache entries in this repository -- `semantic-mutation` (2.72 GiB) and
`rust-native-z3` (1.16-1.24 GiB per platform) -- have each independently been checked for
reducibility and found to reflect what is actually built, not accumulated dead weight
(`docs/DESIGN-ci.md`, "Actions cache budget" and the "Generation coexistence (issue #926,
measured 2026-09-04)" subsection). One lever remains untested: debug-info size.

## Measured (2026-09-04, no `cargo` run)

- `rust/Cargo.toml` pins `z3 = { version = "=0.20.2", default-features = false, features =
  ["vendored", "z3_4_16"] }`. The `vendored` feature compiles the full Z3 C++ solver from source
  as part of the build; this is confirmed as the structural reason `rust-native-z3`'s cache entry
  is large, not a leak.
- **No `[profile]` section exists anywhere in `rust/Cargo.toml` or any crate's `Cargo.toml`**
  (grepped the whole tree). Rust's default `dev`/`test` profile therefore includes full debug info
  (`debug = true`) for every crate, including the vendored Z3 C++ objects and every workspace
  dependency, uncompressed.

## Proposal (unvalidated)

Trim debug info for the cached build -- e.g. `[profile.dev] debug = "line-tables-only"` (or
`false`) workspace-wide, or scoped to just the crates behind the large caches -- to see whether it
measurably shrinks `rust-native-z3` and/or `semantic-mutation`'s Actions cache size.

## Why this needs its own issue instead of a quick change

- A `[profile]` change affects **every** build, not only the cached one, and reduces backtrace
  quality on a CI failure. That is a diagnostics-quality tradeoff that should not be bundled into
  a cache-budget fix (#926) without being named and evaluated on its own.
- Validating this requires an actual `cargo build`/CI run to measure the resulting cache size
  **and** whether it changes cold-build time (stripping debug info could plausibly help, hurt, or
  be neutral for build time; only a measured run settles it, not reasoning from Cargo's docs).
- The two active cargo-using tracks in this repository are currently at their agreed limit
  (#694's full test suite, #698); this does not fit in either until capacity frees up, and is not
  itself authorized to open a third.

## Completion criteria

Before merge, measure and report **both**:
1. The resulting Actions cache size for the affected key(s), compared to the current measured
[...]

PR #985, opened 2026-09-04
fix(ci): judge the cache budget audit's controllable footprint (#926)

### LINKED ISSUE #926: ci(cache): Actions cache budget が main で恒常的に 85% 閾値を超える(削除による緩和は main-cache-absent を生む)
## 観測

`main` の `cache budget audit` が `ca2d5e7d` (push) で failure。run `33047056553`。

```
budget-exhausted: cache usage is 11.55 GiB of a 10.00 GiB limit (116%), at or above the 85%
threshold. 11.55 GiB is 1.55 GiB above the limit; a sufficiently large save can trigger
least-recently-used eviction, including a default-branch cache that a main-targeting pull
request depends on.
```

内訳(実測、上位):

| サイズ | key | 最終アクセス |
|---|---|---|
| 2.72 GiB | `v0-rust-semantic-mutation-Linux-x64-718c915e-9efe1eb7` | 06:29 |
| 2.72 GiB | `v0-rust-semantic-mutation-Linux-x64-0b9fd15e-9efe1eb7` | 06:46 |
| 1.50 GiB | `v0-rust-rust-workspace-Linux-x64-718c915e-9efe1eb7` | 06:46 |
| 1.50 GiB | `v0-rust-rust-workspace-Linux-x64-0b9fd15e-9efe1eb7` | 06:46 |
| 1.35 GiB | `v0-rust-wasm-Linux-x64-718c915e-9efe1eb7` | 06:46 |
| 1.16 GiB | `v0-rust-rust-native-z3-Darwin-arm64-cf75cde1-9efe1eb7` | 06:46 |

`718c915e` 世代と `0b9fd15e` 世代が**並存**しており、大きい3キー
(`semantic-mutation`, `rust-workspace`, `wasm`)がそれぞれ二重に乗ると
それだけで 85% 閾値 (8.5 GiB) に達する。

## 削除による対処を試し、裏目に出た(記録)

旧世代 `718c915e` の3件(2.72 + 1.50 + 1.35 GiB)を削除して 11.55 GiB → 5.99 GiB (60%) にした。
その後 audit を再実行した結果は **failure、finding が 2 件に増えた**:

```
budget-exhausted: cache usage is 8.84 GiB of a 10.00 GiB limit (88%), ...
main-cache-absent: no `refs/heads/main` cache for shared key `wasm` on platform `Linux`.
```

2つの事実が分かった。

1. **使用量はすぐ埋め戻る。** 削除の直後にローカルで測ると 5.99 GiB だが、audit の再実行時点では
   8.84 GiB だった。走行中の CI が保存を続けるため、**この観測量は分単位で動く**。
   単発の計測で「解消した」と判断できない。
2. **削除は `main-cache-absent` を新たに生む。** 現行世代 `0b9fd15e` には `wasm` エントリが無く、
   `wasm` の `refs/heads/main` キャッシュは削除した `718c915e` の1件だけだった。実測で確認:
   削除後に `wasm` を含むキャッシュは `Linux-wasm-bindgen-cli-0.2.126`(8 MB、別キー)のみ。
   結果として main-targeting PR の wasm ジョブがコールドになる。

つまり**削除は運用的な緩和にならない**。削除する場合は最低限「そのキーが `refs/heads/main` の
唯一のエントリではないこと」を事前に確認する必要があり、それを確認すると削除できるキーはほとんど残らない。

## 求める対応(構造的削減)

以下のいずれか、または組み合わせ:

1. **世代の並存を短くする。** キーに含まれる2つのハッシュ(`718c915e` / `0b9fd15e` 側)が何に
   由来するかを明示し、`restore-keys` で旧世代から復元できるなら旧世代の保存を止める。
2. **保存対象を絞る。** `semantic-mutation` (2.72 GiB) と `rust-workspace` (1.50 GiB) が
   同じ `target/` の大部分を二重に持っていないか確認する。共有できるなら1つにする。
3. **共有キーの main エントリを保証する。** `main-cache-absent` が出るキー集合について、
   main で必ず1エントリが保存される経路(どのワークフローがそのキーを save するか)を
   設計文書に書き、監査がその不在を検出したときの復旧手順も書く。
4. `docs/DESIGN-ci.md` の "Actions cache budget" 節 (`docs/DESIGN-ci.md:514`) に、
   **監査が赤になったときの正しい対応**を書く。今回は「削除」を選んで finding を増やした。
   その反例を節に残す。

## 補足

この audit は post-merge の main で走り、PR では `cache budget audit wiring` として別に緑になる。
つまり **PR では検出されず main で初めて落ちる**クラスであり、`native Z3` レーンと同じ性質を持つ。

rizumita:
## Problem

Issue #926: the `cache budget audit` (issue #747) was failing on `main` at 85%+ of the 10 GiB
Actions cache budget, "at least mitigation by deletion produces `main-cache-absent` instead" (the
issue's own recorded experiment).

## Re-measurement, current `origin/main`

The originally-reported cause (`semantic-mutation`/`rust-workspace`/`wasm` holding two coexisting
hash-suffix generations) no longer reproduces. Usage was still 9.07 GiB / 91% on `a3703211`
(2026-09-04), driven instead by `rust-native-z3` Darwin/arm64 holding two generations
simultaneously. Root cause, confirmed by reading both jobs' own `Swatinem/rust-cache` restore
logs (not inferred): `rust/Cargo.lock` and `.github/workflows/ci.yml` are byte-identical across
the whole coexistence window, and the pinned toolchain resolved to the identical rustc commit both
times — what changed was the *runner's ambient installed-toolchain list* that `Swatinem/rust-cache`
also hashes into its key, most likely a `macos-15` runner-image update. This is outside this
repository's control.

## Why this PR does not add a "generation coexistence" detector rule

A candidate rule ("flag 2+ live generations of the same `{key, platform}` on `main` while usage is
≥85%") was designed and rejected: its shortest passing implementation is deleting the older
generation, which is exactly the action the issue's own experiment showed trades
`budget-exhausted` for `main-cache-absent`. A check whose easiest passing implementation is the
action it exists to prevent is not added (condition 4 of this repository's check-addition
discipline).

**The same scrutiny applies to the *existing* `budget-exhausted` check.** Its current form (raw
listing total ≥ 85%) has the identical defect: when the overage is runner-image-driven generation
churn, the shortest way to silence it is still deletion. Recalibrating the threshold alone doesn't
fix this — the clean single-generation footprint (~7.35 GiB, matching the ~7.3–7.5 GiB this
repository already measured and recorded after the #747 incident) plus headroom for the largest key
transiently duplicating (`semantic-mutation`, 2.72 GiB) would push the threshold past the
repository's entire 10 GiB cache ceiling, which would also stop detecting the *original* #747
incident (many concurrent pull-request-scoped caches, not generation churn).

## Change

`auditCacheBudget()`'s `budget-exhausted` rule now judges a **controllable footprint**: for every
`{sharedKey, platform}` pair on `refs/heads/main`, only the most recently created generation
counts; every older generation of the same pair is excluded from judgment (GitHub's own
least-recently-used eviction reclaims it on its own schedule). Every entry on any other ref —
including `refs/pull/*` — is still counted individually and in full, so the original #747 incident
shape is unaffected.

Deleting a stale generation changes nothing this audit judges (verified by a rejecting test that
diffs `findings`/`ok` between "stale generations present" and "already deleted" — only the
informational report differs), removing the incentive to delete.

The raw total and the number of superseded generations excluded are still always reported, labeled
`informational/*`, and never affect `ok`/`findings` — per the lead's explicit requirement that
ambient GitHub-side eviction churn stay visible without being judged.

`docs/DESIGN-ci.md` gained a new "Generation coexistence (issue #926, measured 2026-09-04)"
subsection recording all of the above: the re-measurement, the root-cause log comparison, why the
candidate rule and `shared-key` were both rejected (only one job builds this cache — confirmed by
grepping every workflow — so there is nothing for `shared-key` to consolidate), and why the cache's
size reflects the vendored Z3 C++ build (`z3 = { features = ["vendored", ...] }`) rather than
accumulated debt, mirroring the already-settled `semantic-mutation` case.

## Not done here, deliberately

A further, untested lever remains: no `[profile]` section exists anywhere in `rust/Cargo.toml` or
any crate's `Cargo.toml`, so the vendored Z3 build plausibly carries full, uncompressed debug info.
Filed separately as #984 — validating it needs an actual `cargo`/CI build to measure both cache
size *and* cold-build duration, which this task's cargo-free scope (bug-track-B, running in
parallel with a separate cargo-using track already at its concurrency limit) cannot do.

## Test evidence

`node --test .github/scripts/audit-cache-budget.test.mjs`: 69 passed (65 pre-existing, updated for
the new report/informational shape and re-verified unchanged in substance, plus 4 new). 3
consecutive runs. New calibrated fixtures use the exact real 2026-09-04 `gh cache list` bytes:

- accepting: the measured coexistence (raw 9.07 GiB/91%, controllable ~7.34 GiB/68%) → `ok: true`.
- rejecting (condition-4 proof, both directions): deleting the stale generation changes `findings`/`ok`
  not at all (only `informational`); inflating the *kept* generation alone still trips
  `budget-exhausted` on the de-duplicated total.
- rejecting: de-duplication is scoped to `refs/heads/main` — a pull-request-scoped duplicate of a
  declared shared key still trips the existing rule 3, unaffected by how many generations exist on
  `main`.

Also run and unaffected: `node --test .github/scripts/report-cache-budget-audit.test.mjs` (47
passed), `pytest tests/test_cache_budget_audit_workflow.py` (146 passed),
`.github/scripts/validate-cache-budget-audit-workflow.py` (PASS),
`tools/check-design-citation-headings.py check` (76 citations, 0 findings),
`tools/check_spdx_headers.py`, `tools/aggregate_changelog.sh selftest`/`check-pr` (all PASS). No
`cargo` command was run anywhere in this task.

Closes #926.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/.github/scripts/audit-cache-budget.mjs
+++ b/.github/scripts/audit-cache-budget.mjs
@@ -143,44 +143,46 @@ export function auditCacheBudget({
     return { findings, informational, ok: false };
   }
 
-  // 1. Budget headroom, judged over the *controllable* footprint (issue
-  //    #926): at most one generation per {sharedKey, platform} pair on the
-  //    default branch, keeping only the most recently created generation.
-  //    `Swatinem/rust-cache` hashes the runner's entire installed-toolchain
-  //    list into its key, not only the one version a workflow pins, so a
-  //    `macos-15`/`windows-latest` runner-image update regenerates a key even
-  //    though nothing this repository controls changed (measured
-  //    2026-09-04: `rust/Cargo.lock` and `.github/workflows/ci.yml` were
-  //    byte-identical across a coexistence window, and both runs resolved the
-  //    pinned toolchain to the identical rustc commit). GitHub's own
-  //    least-recently-used eviction already reclaims a superseded generation
-  //    on its own schedule; judging the raw total would report a defect
-  //    nobody in this repository can fix by any means this audit is willing
-  //    to reward, and the shortest way to silence that false alarm --
-  //    deleting the older generation -- has already caused a *worse*,
-  //    previously observed failure (main-cache-absent) for this exact
-  //    reason. See docs/DESIGN-ci.md, "Generation coexistence (issue #926,
-  //    measured 2026-09-04)".
+  // 1. Budget headroom, judged over the *raw, physical* total -- unchanged
+  //    from before issue #926. GitHub's budget and its least-recently-used
+  //    eviction act on physical bytes sitting in the account, not on any
+  //    repository-side classification of which bytes are "controllable";
+  //    "self-healing" describes whether a lever this audit rewards can fix a
+  //    generation, not whether it is currently occupying budget. An earlier
+  //    version of this rule subtracted a de-duplicated {sharedKey, platform}
+  //    total from judgment and was reverted after review found two executed
+  //    counterexamples: (a) two same-identity 5 GiB generations physically
+  //    filling the entire 10 GiB budget were judged as 5 GiB and passed, and
+  //    (b) an independently-observed usage total already higher than the
+  //    listing sum was reduced below threshold by subtracting listing-derived
+  //    stale bytes that are not proven to be the same bytes the usage
+  //    endpoint is counting -- the two observations are non-atomic (see the
+  //    existing max(usageBytes, rawSummed) comment below), so a byte
+  //    identified as stale in one cannot be assumed present, or absent, in
+  //    the other. Both are real physical-budget risks; classifying them
+  //    "controllable" must not make them invisible to `ok`.
   //
-  //    Only default-branch entries attributable to a {sharedKey, platform}
-  //    pair are de-duplicated this way. Every other entry -- including any
-  //    `refs/pull/*` cache -- is counted individually and in full: the
-  //    original #747 incident was many *different* refs each holding their
-  //    own ref-scoped cache, not multiple generations of one key on one ref,
-  //    and de-duplication must not blunt detection of that shape (rules 3
-  //    and 4 below still see every such entry).
+  //    Generation coexistence (issue #926, measured 2026-09-04) is still
+  //    computed and reported below, but strictly as an *additional*
+  //    diagnostic alongside a real `budget-exhausted` finding, never as a
+  //    reduction applied before judgment. See docs/DESIGN-ci.md, "Generation
+  //    coexistence (issue #926, measured 2026-09-04)".
   const rawSummed = caches.reduce((total, entry) => total + (entry.size_in_bytes ?? 0), 0);
   const rawEffective = typeof usageBytes === "number" ? Math.max(usageBytes, rawSummed) : rawSummed;
 
+  // Diagnostic only, computed from the listing alone (the usage endpoint has
+  // no per-entry breakdown to de-duplicate): at most one generation per
+  // {sharedKey, platform} pair on the default branch is "current"; every
+  // older generation in the same group is one `Swatinem/rust-cache` would not
+  // restore today. Every non-main entry -- including any `refs/pull/*` cache
+  // -- is left out of this grouping entirely (not merely "kept whole"): rules
+  // 3 and 4 below already judge those in full, and de-duplication logic must
+  // not touch the shape that caught the original #747 incident.
   const mainGenerationsByIdentity = new Map();
-  let unattributedMainCount = 0;
   for (const entry of caches) {
     if (entry.ref !== defaultBranchRef) continue;
     const { sharedKey, platform } = entryIdentity(entry.key);
-    if (!sharedKey || !platform) {
-      unattributedMainCount += 1;
-      continue;
-    }
+    if (!sharedKey || !platform) continue;
     const identity = `${sharedKey}::${platform}`;
     const group = mainGenerationsByIdentity.get(identity) ?? [];
     group.push(entry);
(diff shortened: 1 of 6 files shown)


rizumita:
## Redesigned after review found two counterexamples (2026-09-04), commit `cc95a907`

The lead reviewed the original design (Sol, independent review) and it was reverted before merge. Summary of what changed:

**The defect (both counterexamples independently reproduced before fixing):**
1. Two generations sharing one `{sharedKey, platform}` pair, 5 GiB each, physically filling the entire 10 GiB budget: the reverted design judged only the newest generation (5 GiB) and passed.
2. An independently-observed `usageBytes` total already at 100% of budget, with listing-derived "stale" bytes subtracted from it: fell back under threshold even though GitHub's own authoritative usage counter was still at capacity. The listing and the usage endpoint are separate, non-atomic observations (already documented via the pre-existing `Math.max(usageBytes, rawSummed)`), so a byte identified as stale in one is not proven to be present, or absent, in the other.

**The fix:** `budget-exhausted` now judges the raw physical total exactly as it did before any #926 change (`Math.max(usageBytes, rawSummed)`, unmodified). Generation-coexistence information is now a strictly **additional diagnostic**, never a judgment input:
- `generation-coexistence-partial-explanation` — a real finding, but present only alongside an actual `budget-exhausted` finding, phrased as "up to N GiB *may be*" self-healing churn (not a subtraction).
- `generation-coexistence` — always present in the non-gating `informational` array, stating the raw entry count and superseded-generation byte count regardless of whether the budget is exhausted.

**Practical consequence:** this audit still reports `budget-exhausted` at 9.07 GiB / 91% for the exact 2026-09-04 measurement in `docs/DESIGN-ci.md`. That is correct — the fix makes the red result actionable (which part is self-healing vs. not), not green.

**Both counterexamples are now permanent regression tests** (`audit-cache-budget.test.mjs`, "review counterexample 1/2"), independently re-run against the reverted design to confirm they actually catch it before being added: [produced/expected below].

**Other findings addressed:**
- `validIsoTimestamp` now requires the exact GitHub timestamp shape plus a round-trip calendar-day match — `Date.parse` alone silently accepted `"0"` and rolled `"2026-02-30"` over to `2026-03-02`; both are now rejecting fixtures in the malformed-field table.
- The de-duplication-scope test now asserts the PR-scoped entry's bytes are **fully present in the reported total** (not merely that the entry exists), proving nothing is silently excluded from judgment.
- `docs/DESIGN-ci.md`'s "checked for reducibility" narrowed to what was actually checked (composition, not every lever) — the same paragraph already discloses the debug-info lever as untested one sentence later.
- Removed the dead `unattributedMainCount` variable.
- The coexistence fixture's comment no longer claims the Windows_NT pair's root cause was independently confirmed the same way Darwin's was (only Darwin's CI logs were diffed).

**Verification (produced/expected, same commit `cc95a907`, rebased onto `origin/main` a5cbc003 post-#946 merge):**
```
$ node --test .github/scripts/audit-cache-budget.test.mjs   # 3 consecutive runs
tests 71
pass 71
fail 0
```
Both counterexamples independently confirmed against the *reverted* (pre-this-commit) design before being kept as permanent tests:
```
counterexample 1 (same-identity 5+5 GiB): reverted design -> budget-exhausted ABSENT (bug); fixed design -> present
counterexample 2 (usage=10GiB, listing 2+2GiB, 1 stale): reverted design -> budget-exhausted ABSENT (bug); fixed design -> present
```
Also re-run and unaffected: `report-cache-budget-audit.test.mjs` (47 passed), `pytest tests/test_cache_budget_audit_workflow.py` (146 passed), `validate-cache-budget-audit-workflow.py`, `check-design-citation-headings.py` (77 citations, 0 findings), `check_spdx_headers.py`, `aggregate_changelog.sh selftest`/`check-pr` (all PASS). No `cargo` command was run anywhere in this task.

Requesting CI on this push to complete before returning this PR to ready-for-review; the lead's merge decision remains separate.

Issue #1020, opened 2026-09-09
strict-tags: native Declared side never collects requirement-block IDs, so an unformalized empty requirement block is never reported


rizumita:
## What the design document declares

`docs/DESIGN-strict-tags.md` section 2, the `Declared` bullet:

> **Declared** = `--requirements ids.txt` (one ID per line) ∪ requirement-block IDs of the
> requirements dialect (auto-collected via `__requirement_ids` — this also catches an
> **empty requirement block** "declared but forgotten to formalize." Collection is essential
> because the trace disappears after expansion).

## What the native implementation executes

At `feedaefa`, the whole declared side of `strict_tag_warnings_from_source` is
`rust/fslc/src/main.rs:5971-5993`:

```rust
    if let Some(path) = requirements {
        let source = std::fs::read_to_string(path)…?;
        for requirement in source.lines().map(str::trim).filter(|line| !line.is_empty()) {
            if !referenced.contains(requirement) {
                warnings.push(json!({ "kind": "unreferenced_requirement", … }));
            }
        }
    }
```

The lines of the `--requirements` file are the entirety of `Declared`. There is no collection
corresponding to `__requirement_ids`, and without the flag the declared side does not run at all.

The frozen Python reference does collect them — `src/fslc/model.py:1479-1480` at the same commit:

```python
    declared = set(spec.get("requirement_ids") or [])
    declared.update(requirement_ids or [])
```

i.e. the spec-side block IDs first, then the file's IDs unioned onto them.

## Measured

Binary: built in this session from `f81c4065` (one commit above `feedaefa`; the fix for #1001,
which does not touch the declared side), `cargo build -p fslc-rust --bin fslc`, sha256
`ebd1ff0c2f5b4d6d8aaeddae43c19e5c7c1f4f2671091e4e009136a4c6c2ab4f`, `worktree_fp` equal before
and after the build, `dirty_tracked_files=0`. Not a binary found in `rust/target`.

Fixture — a `requirements` spec with two referenced IDs and one **empty** requirement block:

```fsl
requirements EmptyRequirementBlockDeclared {
  state { n: 0..3 }
  @requirement("REF-001", "a referenced requirement")
  init { n = 0 }
  @requirement("REF-002", "a referenced requirement on an action")
  action bump() { requires n < 3
    n = n + 1 }
  requirement REQ-EMPTY "declared but never formalized" {
  }
}
```

| # | invocation | expected per the document | produced |
|---|---|---|---|
[...]

PR #1024, opened 2026-09-09
fix(traceability): count init and trace-case annotations as referenced (#1001)

### LINKED ISSUE #1001: fix(traceability): forbidden直前の@requirementを未参照と誤診断する
## Problem

`@requirement(...)` を `forbidden` ブロックへ直接付けても、
`fslc check --requirements` がそのIDを
`unreferenced_requirement` と報告する。

警告のhint自身は「declaration tag, acceptance, or forbidden block」を参照元として挙げるため、
実際のソースと診断が矛盾している。

## Environment

- release: `fslc 4.4.1`
- latest main: `dbbf6889a0c1f3001552952ffc537632a01f8188` (also reproduced)
- observed on: macOS arm64

## Minimal reproduction

`forbidden-traceability.fsl`:

```fsl
requirements ForbiddenTraceability {
  enum Performer { Human, AI }

  state {
    approved: Bool
  }

  init {
    approved = false
  }

  requirement REQ-TST-001 "Only a human can approve" {
    action approve(performer: Performer) {
      requires performer == Human
      approved = true
    }
  }

  @requirement("ASSUME-TST-001", "AI approval is forbidden")
  forbidden FB-TST-001 "AI cannot approve" {
    approve(AI)
    expect rejected
  }
}
```

`requirements.txt`:

```text
REQ-TST-001
ASSUME-TST-001
```

```console
$ fslc check forbidden-traceability.fsl --strict-tags --requirements requirements.txt
{
  "result": "ok",
  "warnings": [
    {
      "kind": "unreferenced_requirement",
      "name": "ASSUME-TST-001",
      "hint": "no declaration tag, acceptance, or forbidden block references this requirement ID"
    }
  ]
}
$ echo $?
0
```

The forbidden scenario itself is valid: its final `approve(AI)` step is rejected by the
`performer == Human` precondition.

## Expected

`ASSUME-TST-001` is counted as referenced by `FB-TST-001`, so the warning is absent.

## Acceptance criteria

- [ ] Requirement annotations on a `forbidden` block participate in external-ID traceability.
- [ ] The equivalent `acceptance`, invariant/trans, and requirement-declaration paths remain covered.
- [ ] Removing the annotation is a rejecting control and restores `unreferenced_requirement`.
- [ ] The native CLI test pins the warning projection and exit code.

rizumita:
Closes #1001.

## Problem

`fslc check|verify --strict-tags --requirements <ids>` reports a requirement ID as
`unreferenced_requirement` even when a typed `@requirement(...)` annotation links a declaration to
it. The warning's own hint names the sources it consults —

> no declaration tag, acceptance, or forbidden block references this requirement ID

— so the diagnostic and the source contradict each other, which is what #1001 reports for
`forbidden`.

`acceptance` and `init` have the same defect. The issue's second acceptance criterion says the
"equivalent `acceptance` … paths **remain covered**", which presupposes they already work; measured,
they do not. A correction with the measurement is recorded on #1001 itself.

## Which surface forms are affected, and why that is the population

`docs/LANGUAGE.md:2035-2043` names the declaration-level attachment points; `:1930-1931` says a
`requirement`/`acceptance`/`forbidden`/… declaration "owns the ID after its keyword" while
`@requirement(...)` "links another declaration to such an ID". So both the owned and the linked ID
are referenced IDs — the requirement is stated about **surface forms**.

The map from surface form to storage is not guesswork: `docs/DESIGN-annotations.md:116-118` states
that attachment "reuses the existing stable target vocabulary (`action_target`, `property_target`,
`INIT_TARGET`) — no new target-key functions were needed", and the map itself is implemented in one
function, `model.rs::collect_declaration_annotations` (`:1357-1486`), whose final arm
(`_ => (None, …)` at `:1469`, with `let Some(target) = target else { continue }` at `:1471`) drops
every variant not listed above it.

| # | surface form | binds to | evidence | counted before this change? |
|---|---|---|---|---|
| 1 | `init` | `INIT_TARGET` → `KernelModel.init_annotations` | `model.rs:1373`, `:965` | **no** |
| 2 | `action` (incl. sync, mapped/requirement) | `action_target(name)` | `model.rs:1385` | yes |
| 3 | `invariant` | `property_target("invariant", name)` | `model.rs:1397` | yes |
| 4 | `trans` | `property_target("trans", name)` | `model.rs:1402-1416` | yes |
| 5 | `reachable` | `property_target("reachable", name)` | `model.rs:1428` | yes |
| 6 | `until` | **both** `property_target("trans", "{name}_until_safety")` and `property_target("leadsTo", name)` | `model.rs:1440-1445` | yes |
| 7 | `unless` | `property_target("trans", name)` — shares the `Trans` match arm | `model.rs:1409-1416` | yes |
| 8 | `leadsTo` | `property_target("leadsTo", name)` | `model.rs:1464` | yes |
| 9 | process `transition` (`covers`) | `action_target(&transition.name)` | `dialect.rs:1171-1181` | yes, via #2 |
| 10 | `requirement` block | fans out to every contained action/property target | `LANGUAGE.md:2035-2036`, `DESIGN-annotations.md:121-124` **[1]** | yes, via #2/#3 |
| 11 | `acceptance` block | `RequirementsTraceCase.annotations` | `dialect.rs:3141-3150` | **no** |
| 12 | `forbidden` block | `RequirementsTraceCase.annotations` | `dialect.rs:3162-3170` | **no** |

**[1]** Row 10's evidence is weaker than every other row's. Rows 1-9, 11 and 12 cite the line that
performs the binding; row 10 cites the two prose sentences that say the fan-out happens, and the
site that performs it is not identified here.

Three forms are not counted (1, 11, 12), landing on **two** collection sites — 11 and 12 share one,
which is why the site count is smaller than the form count.

## Contract change — two sites, fixed two different ways

The behavioural change is small and lives in `rust/fslc/src/main.rs`. Its two halves are not the
same kind of edit and should not be read as one:

**1. An addition to an enumeration.** `strict_tag_warnings_from_source` chains the annotation sets
of `actions`, `invariants`, `transitions`, `leadstos` and `reachables`; `init_annotations` was
missing. One `.chain(std::iter::once(&model.init_annotations))` adds it. This is still an
enumeration, and a future annotation-bearing field can still be forgotten here.

**2. The removal of an enumeration.** The trace-contract branch read

```rust
referenced.extend(contract.acceptance.into_iter().map(|case| case.id));
referenced.extend(contract.forbidden.into_iter().map(|case| case.id));
```

— two lines, each naming one of the two things a case can contribute, and neither reaching the
annotations. `dialect.rs::trace_case_annotations` (`:3197-3213`) builds a case's `Annotations` by
synthesising `Annotation::Requirement { id: case.id, text }` **first** and then extending with the
surface annotations, so `case.annotations.requirements()` yields the owned ID *and* every linked ID.
Reading that instead of `case.id` makes the two lines one loop and removes the shape in which one of
the two can be forgotten. `docs/DESIGN-annotations.md:124-127` states this as the accepted design,
naming the same function: "acceptance/forbidden block annotations extend the trace case's own
`Requirement{id,text}` relation (`dialect.rs::trace_case_annotations`)". The producing side already
implemented it; only this consumer discarded it.

**Nothing in `fsl-core` changes.** The defect was entirely in the consumer.

**A third edit is not part of the fix.** The referenced-ID construction now lives in
`referenced_requirement_ids`, because adding to the chain pushed
`strict_tag_warnings_from_source` to 104 lines and `clippy::too_many_lines` (threshold 100) failed
for real. Extracting is the honest way past that. The two alternatives were rejected on principle:
compressing the lines would make the shortest way to pass a line counter diverge from the way that
satisfies the requirement, and `#[allow(clippy::too_many_lines)]` is the suppression flag
`AGENTS.md` forbids. Afterwards the two functions are 36 and 83 lines. The extracted function's doc
comment states the same asymmetry as this section — that its first half is an enumeration a future
field can fall out of, and its second is not.

**The hint text is unchanged. That is a judgement, recorded below, not an oversight** — and the
first version of this section asserted the opposite for the wrong reason. See "The hint, judged".

## Test evidence

One test is added, `strict_tag_traceability_counts_init_block_and_trace_case_annotations`, holding
two controls over two fixtures. Both are labelled by what they establish, and both labels were
earned by running the mutation rather than asserted. **The name states the sites the change adds**;
an earlier name said "every documented reference source", which quantified a population the fixture
does not cover — seven of the twelve forms named above carry annotations in it.

**The annotated fixture is a detector.** It carries `@requirement(...)` on seven attachment points —
`init`, an action, an invariant, a `trans`, a `requirement` block, an `acceptance` block and a
`forbidden` block — plus one ID (`MISSING-001`) that nothing references. The case IDs are distinct
from the annotation IDs (`AC-TRACE-001` vs `ACCEPT-001`), so counting case IDs cannot accidentally
satisfy the assertion. The whole `warnings` array is compared, for `check` and for success-path
`verify`, and the exit code is pinned separately.

Calibrated by removing the fix — `main.rs` reverted to the merge base's version, which the gate
recorded as `dirty_tracked_files=2` (the file went clean) with `worktree_fp` equal before and after:

| | value |
|---|---|
| produced (pre-fix) | `INIT-001`, `ACCEPT-001`, `FORBID-001`, `MISSING-001` |
| expected (post-fix) | `MISSING-001` |
| process exit | 101, panicking at `cli_regression.rs:1034` |

Comparing the two ID sets in **both** directions, rather than their sizes: `before − after` is
exactly `{INIT-001, ACCEPT-001, FORBID-001}`, and `after − before` is empty. Equal counts would not
have shown that the three that disappeared are the three intended, nor that no new warning appeared.

**The unannotated fixture is a rejecting control**, not a second detector. It is the same
specification with the seven annotations removed and the case IDs kept, so the fix cannot add
anything to the referenced set. Of the ten IDs the shared requirements file declares, its array must
still list the **eight** annotation IDs as unreferenced — the two trace-case IDs stay referenced
there, because lowering synthesises a case's own ID whether or not the case carries annotations, and
that is precisely what the control in the next paragraph is for. Measured on both sides of the
change: 11 warnings (3 `untagged` for `bump`/`ReadyBounded`/`Monotone`, then the 8
`unreferenced_requirement`), identical before and after. This is what fails if the collection ever
becomes over-broad: a fix that marked everything referenced would empty it.

**The trace cases' own IDs are declared too, and that is the control for the half of the fix that
*replaced* a mechanism.** `strict_tag_traceability_requirements.txt` lists `AC-TRACE-001` and
`FB-TRACE-001` alongside the eight annotation IDs. Without them the replacement had no control at
all: the fixture declared only the annotation IDs, so losing the own-ID synthesis in
`dialect.rs::trace_case_annotations` would have failed nothing while the commit message claimed the
own ID survives. Adding them changes **neither** expectation — measured, not assumed
(`test result: ok. 1 passed`) — and the control was calibrated by deleting the synthesis at
`dialect.rs:3203-3207`:

| | value |
|---|---|
| exit | 101, panicking at `cli_regression.rs:1034:9` |
| produced | `AC-TRACE-001`, `FB-TRACE-001`, `MISSING-001` |
| expected | `MISSING-001` |
| **surfaced by the mutation** | **exactly the two trace-case IDs** — not all four, so the mutation is isolated to that synthesis and nothing else |

Restored by `git show HEAD:… >`, verified by sha256 equality and an empty `git diff HEAD`, and green
again afterwards.

⚠️ **This calibration was outstanding when the PR was opened, and has since been taken at this
SHA.** It had first been executed against the same file contents but at an earlier `worktree_fp`,
before the commit was amended to its final form, and re-running it here was refused by the local
resource gate 61 times while another worktree held the build window. Identical content is not the
same measurement, so it was named as outstanding rather than reported as done. Taken at
`6a2376f7`: exit 101 at `cli_regression.rs:1034:9`, produced `AC-TRACE-001`/`FB-TRACE-001`/
`MISSING-001` against expected `MISSING-001` — **surfaced exactly the two trace-case IDs** — then
restored by sha256 equality and green again. The numbers are in a comment on this PR. **Every gate
and calibration cited in this body was taken at this commit.**

**The rule the amended `Referenced` bullet states was measured against three input classes.** The
binary was built in this session from this commit (`cargo build -p fslc-rust --bin fslc`, `dirty=0`,
`worktree_fp` equal before and after); a binary already sitting in `rust/target` is not evidence
about the tree, and neither is the installed `~/.local/bin/fslc`. Each input was run twice,
sequentially, and the two runs are identical in the **whole envelope**, not merely in the warning
names:

| input | how IDs are attached | expected | produced |
|---|---|---|---|
| typed only | `@requirement(...)` ×7 | `MISSING-001` only | `MISSING-001` only |
| legacy only | the `"ID: text"` slot ×2 | `REQ-NEVER-USED` only | `REQ-NEVER-USED` only |
| **both, on one declaration** | `TYPED-001`/`-002` **and** `LEGACY-001`/`-002` | `NEVER-USED` only | `NEVER-USED` only |

All three exit 0 with `result: "ok"`. Each input carries its own rejecting side — the one declared ID
in neither channel is warned — so a build that emitted nothing at all would not pass.

**The third input is the one that carries the claim, and its two channels name different IDs on
purpose.** Had they named the same ID, the legacy slot alone would make the declaration referenced
and the typed channel's contribution could never appear in the envelope. Because they differ, each
channel's collection is the only thing keeping its own two IDs out of the warning list — which makes
the input falsifiable, and it was falsified, one channel at a time:

| mutation (`fsl-core/src/model.rs`, `collect_declaration_annotations`) | expected | produced |
|---|---|---|
| remove the typed `extend` | `NEVER-USED` **+ `TYPED-001`, `TYPED-002`** | `NEVER-USED`, `TYPED-001`, `TYPED-002` |
| remove the legacy `bind` | `NEVER-USED` **+ `LEGACY-001`, `LEGACY-002`** | `LEGACY-001`, `LEGACY-002`, `NEVER-USED` |
| neither | `NEVER-USED` only | `NEVER-USED` only |

Neither mutation surfaced all four IDs, which is what shows they are isolated rather than one
compound change. Each was reverted with an sha256 equality check against the recorded baseline and
an empty `git diff HEAD`. So the union in the `Referenced` bullet is **observed inside one envelope
and shown to break in both directions**, not inferred from the first two rows.

**What the two preservation controls preserve, and when that preservation breaks** — stated because
a control that cannot break carries no information:

| control | preserves | breaks when |
|---|---|---|
| the unannotated fixture | that removing every annotation leaves all eight IDs unreferenced | the collection becomes over-broad — a fix that marked everything referenced would empty its array |
| legacy-only input | that the legacy `"ID: text"` channel is still counted after the refactor | the legacy `bind` stops reaching the target (executed above) |
| both-channels input | that the two channels union on one declaration | either channel's collection is dropped (executed above, one at a time) |

The legacy-only and both-channels inputs are **preservation controls with respect to this fix**, not
detectors of it: measured against a binary built with `main.rs` reverted to the base commit, both
produce a **byte-identical whole envelope**. Only the annotated fixture changes across the fix. That
distinction is what keeps them from being presented as coverage they do not provide.

**Gates, all at this commit, with `worktree_fp` equal before and after and no uncommitted tracked
file.** Test counts are given so a silent skip would be visible:

| gate | exit | collected |
|---|---|---|
| `cargo fmt --all -- --check` | 0 | — |
| `cargo clippy -p fslc-rust --all-targets --locked -- -D warnings` | 0 | — |
| focused `--test cli_regression strict_tag_traceability` ×2 | 0 / 0 | `1 passed; 34 filtered out` each |
| `--test cli_regression` ×2 | 0 / 0 | `35 passed` each — the 34 that existed plus this one |
| `--test injection_detector_matrix` | 0 | see below |
| `aggregate_changelog.sh check` / `check-pr` | 0 / 0 | `BASE_SHA=feedaefa HEAD_SHA=<this commit>` |

⚠️ Two exit codes appearing in this session's logs are **not** results and are not counted above:
`8` is the resource gate refusing to start (the gate did not execute — 15 such refusals were
absorbed by retrying while another worktree held the build window) and `143` is a `SIGTERM` from
stopping a ladder that an amend had superseded.

`injection_detector_matrix` is run even though the injected corpus carries no `@requirement` at all
across its 21 files, so `MATRIX.json` cannot move. "Zero occurrences" is a quantified claim that goes
false the first time an annotated spec is added to that corpus, and a claim in a review comment does
not notice when that happens.

The changelog gate was itself calibrated: renaming the fragment's category to the undeclared `fix`
makes `check` exit 1 with `changelog-fragment-name-invalid:
1001-forbidden-annotation-traceability.fix.md` — it names the file, so the gate is reading this
change rather than merely passing near it.

## Documentation

`docs/DESIGN-strict-tags.md` section 2 is amended in four places. Line numbers are at this commit
unless a base-commit number is named, because three of these lines move.

**The `Referenced` definition (`:36-53`; `:27` at the base commit).** It read "the `meta.id` of
every element ∪ acceptance IDs ∪ forbidden IDs", where "element" was that section's own item-1 list.
`meta.id` predates typed annotations, and `docs/DESIGN-annotations.md:116-127` is the accepted
design that replaced it. It now states the rule — a requirement ID is referenced by appearing in the
annotation set of a declaration whose annotations lowering binds to a target key, plus every trace
case's own set — rather than by listing which forms qualify, so a new attachment point is covered
without editing the sentence. **The word "typed" is deliberately absent**: the legacy `"ID: text"`
slot binds to the same target and is counted, which is measured in Test evidence rather than
assumed. Two limits are stated beside the rule because neither follows from it.

**Item 2's headline (`:28-29`).** It read "declared requirement IDs that appear in no declaration
tag, acceptance ID, or forbidden ID", which the amended bullet beneath it then contradicted; it now
reads "no declaration and no trace case references". ⚠️ **That headline is not the shipped hint's
wording** — the hint says "acceptance, or forbidden **block**" — so correcting it does not touch the
message contract deferred to #1017. One consequence is recorded rather than left implicit: **after
this change the document no longer carries the shipped hint verbatim anywhere.** The hint string
itself is untouched, and #1017 is where it is tracked.

**Item 1, the `untagged` bullet (`:17-27`) — raised by independent review, and false rather than
merely stale.** It listed "user invariants / actions / leadsTo / reachable" and gave the criterion as
"no `meta`". Both parts are wrong. `trans` is scanned too, by the native implementation
(`rust/fslc/src/main.rs:5966`) **and** by the frozen Python reference (`src/fslc/model.py:1464`), so
that omission is not a native-versus-Python question. The criterion is, and it took **three** passes
to state without introducing a new falsehood — recorded here because the sequence is the useful part:

| pass | wording | why it was wrong |
|---|---|---|
| original | "have no `meta`" | native's criterion is the annotation set, not the legacy slot |
| 1st fix | "whose annotation set is empty … both channels reach the set" | true of native, written **unmarked** in a document whose section-2 heading is the Python signature; the frozen reference asks about `meta` alone (`src/fslc/model.py:1446`) |
| 2nd fix | "carry no requirement metadata" — chosen to be true of both | **false for native**: `Annotations::source_order` returns every variant (`rust/fsl-syntax/src/annotation.rs:166-185`, `:320-322`), so `@undecided` or `@kind` alone makes a declaration tagged with no *requirement* metadata present |
| 3rd fix (shipped) | leads with "that nothing tags", then states each implementation's criterion verbatim and names the counterexample | — |

⚠️ **Neutralising a one-sided claim is not avoiding a claim; it is making a new one**, and the
neutral phrase needed the same checking the specific one got. The counterexample is measured, not
reasoned: the `@undecided`-only invariant in
`rust/fsl-tools/tests/fixtures/document_claims_fixture.fsl:60-63` produces **no** `untagged`
warning. Only the false parts are changed; the bullet is not restructured. It is fixed here rather
than deferred because amending `Referenced` made `:36` state the native contract, and leaving item 1
false a few lines above it would reproduce exactly the intra-item contradiction the headline had.

**The `Declared` bullet (`:30-35`) is marked, not rewritten.** It claims the `--requirements` file is
unioned with requirement-block IDs auto-collected via `__requirement_ids`, which "also catches an
empty requirement block". Native does not do that: the file's lines are the whole declared side
(`rust/fslc/src/main.rs:5971-5993` **at the base commit**), while the frozen Python reference does
collect them (`src/fslc/model.py:1479-1480`). Measured four ways — see Residues. The bullet describes
the frozen Python surface that #996 is open for, so it receives one clause saying native does not
implement the auto-collection, and the divergence is filed as **#1020**. Marking is not a weakening;
leaving it unmarked is what permits reading it as native behaviour. The clause this change had
originally added beside it — "which is why the Declared side collects block IDs separately" — is
removed, because it leaned on behaviour native does not execute.

`docs/LANGUAGE.md`, `docs/LANGUAGE.ja.md` and `skills/` are **not** changed. They already state the
rule correctly; this was an enforcement gap, not a documentation gap — the shape milestone M19 is
about. `tools/build_site_reference.py` regenerates `docs/intro/{language,cli}.{ja,en}.html` and the
tree is unchanged afterwards, which is what
`.github/workflows/site-reference-freshness.yml` asserts.

## The hint, judged

The warning's hint is left exactly as it is. The reasoning is recorded because the question is not
obvious and the first attempt at it was wrong.

**What "declaration tag" means in that sentence.** The hint at `rust/fslc/src/main.rs:6014` was
almost verbatim from that document's item-2 headline, which **this change revises** (see
Documentation). The document still defines the term at `:3-4`:
"Declaration tags `"REQ-n: original text"` are implemented but optional." So in the hint's own
source vocabulary, a *declaration tag* is the **legacy string slot**, not a typed
`@requirement(...)` annotation.

That is worth stating plainly because the same spelling means something wider elsewhere:
`docs/LANGUAGE.md:1915` titles the section that documents typed annotations "### 13.1 Declaration
tags (traceability common to all layers)". Reading the hint with LANGUAGE.md's sense and concluding
it already covers `init` is exactly the mistake this section replaced.

**So the hint under-describes the behaviour — and did so before this change.** Under the narrow
sense, it never named the typed annotations on `action`, `invariant`, `trans`, `reachable`, `until`,
`unless`, `leadsTo`, a process `transition`, or a `requirement` block, all of which were already
counted. This change adds `init` to that same under-described set.

**What changes is the direction of the inaccuracy, and that is the part that mattered.** Before, the
hint *over-promised*: a user who read it and annotated a `forbidden` block still got the warning —
which is #1001. After, that over-promise is gone; what remains is an under-description, and a user
who follows the hint is not led anywhere false.

**The test this PR adds does pin the hint verbatim, and that is deliberate.** Leaving the hint
alone while pinning it is two decisions, so both are stated.

- The issue's fourth acceptance criterion is "The native CLI test pins **the warning projection** and
  exit code." `hint` is a field of that projection, alongside `kind`, `element`, `name` and `loc`, so
  pinning it is what that criterion asks for.
- The hint is part of the JSON envelope, which `AGENTS.md` requires the native CLI to preserve. A
  shipped message changing should fail a test: that is how such a change becomes deliberate instead
  of silent, and silence is the structure that let #1001 sit unnoticed. The cost is a one-line
  update, and it is already written down — this test is row 2 of the table below, and #1017 lists it
  as one of the surfaces any hint change has to move.

Pinning it adds no pressure toward a wrong implementation: the `hint` field is identical before and
after this change, so the shortest way to make the test pass is still to collect the annotations.

**What every surface pinning the hint would cost, if it moved** (enumerated before deciding, not
after):

| # | surface | kind |
|---|---|---|
| 1 | `rust/fslc/src/main.rs:6014` | the producer (authoritative Rust) |
| 2 | `rust/fslc/tests/cli_regression.rs:942` | **the test this PR adds.** No pre-existing Rust test pins the string |
| 3 | `src/fslc/model.py:1487` | the frozen Python twin carries the same sentence |
| 4 | `docs/DESIGN-strict-tags.md` | where the wording came from. ⚠️ **After this change the document no longer carries it verbatim** — the item-2 headline that did is revised, for the reason given in Documentation |

Row 3 is the reason this is not folded in: rewording the Rust hint alone diverges the two
implementations' text, and `AGENTS.md` allows changing the frozen reference only when an accepted
compatibility decision requires both to move. That is a separate decision from this bug fix.

**The document's item-2 headline was first left alone for the same reason, and that judgement was
withdrawn on measurement.** The initial reading was that `:28-29` *is* the shipped hint's wording, so
changing it would be the same decision as changing the hint. It is not: the hint says "acceptance, or
forbidden **block**", the headline said "acceptance ID, or forbidden ID". Once `Referenced`
(`docs/DESIGN-strict-tags.md:36-53`) was amended to state the native contract, the unrevised
headline contradicted the bullet directly beneath it, and a
document that contradicts itself inside one item is worse than one whose phrasing is merely older. So
the headline is revised and **the hint is not**. The two are separable precisely because they were
never the same sentence.

**The strongest evidence that the wording is stale needs no document at all: the same command's two
hints disagree with each other.** The `untagged` hint tells the user to "tag the declaration with a
typed annotation such as `@requirement(...)`" and adds that "the legacy `"REQ-1: text"` string slot
is non-canonical (`fslc lint` reports it as `legacy_string_metadata`)" — while the
`unreferenced_requirement` hint uses that same legacy form's name as one of the sources it consults.

**Filed as #1017**, so the decision does not live only in this PR body. It records the four pinning
surfaces, the frozen-Python coupling that makes the wording a compatibility decision rather than a
wording fix, and the severity as observed: an under-description, not a false green — the hint does
not participate in any verdict.

The result is that `docs/DESIGN-strict-tags.md` still holds more than one state at once,
deliberately. Line numbers at this commit:

| line / section of `docs/DESIGN-strict-tags.md` | describes | why it is in that state |
|---|---|---|
| `:17-27` (`untagged`), `:28-29` (item-2 headline), `:36-53` (`Referenced`) | the authoritative Rust contract, as amended here | a superseded definition plus two statements that were false against both implementations; basis inside `DESIGN-annotations.md` and the implementation lines cited above |
| `:30-35` (`Declared`) | the frozen Python surface, **now marked** | #996's structural change; native's divergence marked in place and filed as #1020 rather than rewritten here |
| `## 3` (`:61-71`) | the frozen Python surface, **unmarked** | its exclusion mechanism is `__generated`/`generated_names`; native excludes by the `_` name prefix (`rust/fslc/src/main.rs:5983`). Same class as `:30-35`, but in a different section and outside this change's item — recorded here, not fixed |
| `## 4` (`:73-79`), `## 5` (`:81-87`) | the frozen Python surface, unmarked | `model.py`/`cli.py`/`tests/test_strict_tags.py`; the structural change #996 is open for |

⚠️ The document no longer carries the shipped hint verbatim, which the previous version of this
section relied on. #1017 is the record for the hint's wording.

## Residues, stated rather than left implicit

- **Forms 6, 7, 8 and 9 have no fixture in this change.** Their verdict in the table above is read
  off the cited implementation lines, not observed in a `warnings` array. "The design says so and
  the implementation lines say so" is not "we observed it", and this PR does not claim to have
  covered all twelve forms.
- **Row 10's evidence is prose, not a binding site**, as its footnote says.
- **`docs/DESIGN-strict-tags.md` is left holding three states at once**, listed in the table at the
  end of "The hint, judged" rather than restated here. The short form: `## 4. Ripple` (`:73-79`) and
  `## 5. Tests` (`:81-87`) still describe the frozen Python surface, which is the structural change
  #996 is open for. This PR is recorded there as that issue's second instance
  (`#996#issuecomment-5580149281`) instead of being fixed opportunistically here.
- **The hint's under-description survives this PR**, with the enumeration of what moving it would
  cost and why it is a separate decision — again in "The hint, judged".
- **The native declared side does not collect requirement-block IDs**, so an *empty* — declared but
  unformalized — `requirement` block is never reported unless the user separately lists its ID in a
  `--requirements` file. That is the case the `Declared` bullet names as its own reason to exist.
  Measured four ways (the flag alone; the file listing other IDs; the file listing that ID, which
  **does** warn and so separates the missing collection from a broken reconciliation; and a
  non-empty block, which is referenced through its contents as documented). Direction: **fail-open**
  — a missing warning in an opt-in lint, not a verdict, so not a soundness defect. Filed as
  **#1020**; the bullet is marked in place rather than rewritten, because it describes the frozen
  Python surface #996 is open for.
- **`docs/DESIGN-strict-tags.md` `## 3` (`:61-71`) is the same class as `Declared`, left unmarked
  and unfixed.** Its exclusion mechanism is `__generated`/`generated_names`, a Python construct;
  native excludes by the `_` name prefix (`rust/fslc/src/main.rs:5983`). It is in a different
  section and outside this change's item, so it is recorded here rather than edited.
- **The frozen Python reference diverges from the shipped behaviour on both criteria in section 2**,
  and that is not filed: `docs/DESIGN-annotations.md:129-136` already records that the frozen oracle
  has no annotation concept. **Whether anything records the divergence** is the test this change
  writes into the `Referenced` bullet, and it is why #1020 exists and this does not.
- **How much to trust the sweeps.** Their populations had to be repaired **three** times, each time
  because a count looked wrong rather than because the text did: `\b…\b` does not fire inside
  `snake_case` (a sweep over a file containing `..._counts_every_documented_...` returned zero
  hits); bare `` `:NN` `` citations were missed because their filename sits on an earlier line
  (nine of this body's citations were wrong, and the mechanical check had found none of them); and
  the unit was a line while a claim spans lines (so a correctly placed marker still produced 12
  hits). All 25 citations in an earlier draft **resolved** to a real, non-empty line — and nine of
  them still did not support the claim beside them. **Existence is not verification.**

## Review

Independent review ran **four rounds** against a declared stopping rule of zero findings, each on
the commit as it then stood, each returning severity-labelled findings with `path:line` evidence and
an explicit "not read" section:

| round | commit | result |
|---|---|---|
| 1 | `f81c4065` | blocker 0 / major 0 / **minor 2** — the `untagged` bullet false against both implementations; the test name quantified coverage the fixture does not provide |
| 2 | `b74476a7` | blocker 0 / major 0 / **minor 1** — the first correction of that bullet stated native's criterion unmarked |
| 3 | `d879043c` | blocker 0 / major 0 / **minor 2** — the neutral phrasing was itself false for native; the replaced `case.id` mechanism had no control |
| 4 | `6a2376f7` | **blocker 0 / major 0 / minor 0** |

Between rounds 2 and 3 the orchestrator's own sweep of the change's added lines found a third
instance of round 2's class — the `Referenced` bullet stating the native contract unmarked while its
two neighbours had just been marked — which is why the markers are now consistent across the whole
of section 2 rather than bullet by bullet.

⚠️ **What is and is not independent here**, stated as a map rather than as a disclaimer, because
"not independent" understates part of it and overstates another:

| step | series |
|---|---|
| requirement, acceptance conditions, rulings | Claude |
| implementation | Cursor `composer-2.5` |
| review, four rounds | Sol (GPT) |
| final cross-series read | Claude |

So the implementation and the review **are** different series from each other and from the
requirement. What is **not** covered is that the series which wrote the requirement is also the one
that approved it, and that the brief, the perspectives and the acceptance conditions for all four
review rounds were written by this change's own author — disclosed in each brief. What the rounds
establish is that four passes at those perspectives no longer find anything; they do not establish
that the perspectives were the right ones.

---


rizumita:
## The outstanding calibration, now taken at this SHA

The PR body names one measurement as outstanding: the own-ID control's calibration had been run
against the same file contents, but at an earlier `worktree_fp`, and re-running it at this commit
kept hitting the local resource gate's refusal while another worktree held the build window.

It has now been taken at `6a2376f7`, with a clean tree before and after.

| | value |
|---|---|
| mutation | the case's own `Requirement{id,text}` synthesis deleted from `dialect.rs::trace_case_annotations` (`dialect.rs:3203-3207`) |
| exit | **101** |
| panic | `fslc/tests/cli_regression.rs:1034:9`, in `strict_tag_traceability_counts_init_block_and_trace_case_annotations` |
| collected | `FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 34 filtered out` |
| produced | `AC-TRACE-001`, `FB-TRACE-001`, `MISSING-001` |
| expected | `MISSING-001` |
| **surfaced by the mutation** | **exactly `AC-TRACE-001` and `FB-TRACE-001`** — two, not four |
| restore | `git show HEAD:… >`, sha256 equal to `df8e251b…`, `git diff HEAD` empty, `worktree_fp` back to its pre-mutation value |
| green after restore | exit **0**, `test result: ok. 1 passed; 0 failed; 34 filtered out` |

**Why the "two, not four" line is the point.** The mutation had to fail *for the reason it is cited
for*. Had all four declared IDs surfaced, the mutation would have broken more than the own-ID
synthesis, and the control would not be evidence about that synthesis specifically. Only the two
trace-case IDs appear, and the eight annotation IDs stay referenced — so what the control detects
is exactly the property the change's second half replaced a mechanism to preserve.

With this, **every gate and calibration cited in the PR body was taken at this commit.** Nothing is
carried over from an earlier tree state.

Two notes on how it was measured, since both bit this change earlier:

- The local resource gate refused to start (`exit 8` — the gate did not execute, which is neither a
  pass nor a failure) **61 times** before the own-ID mutation step ran, and **20 times** before the
  green-after step ran: **81** refusals in total. None of them is counted as a result anywhere.
  *(Corrected: this line first said "62 times", a figure taken from a log's total refusal count
  rather than from the attempts for these two steps. Counted, it is 61 + 20.)*
- The first re-take script applied the mutation and *then* waited for the window, leaving the
  working tree modified for about thirty minutes while it retried. That was rebuilt so the wait
  happens with a clean tree and the mutation is applied only once the window is open; if the budget
  runs out it now exits without ever touching the tree.

Issue #1043, opened 2026-09-15
diff(outcome): `outcome_class` の `semantic_diff` が top-level `violations` を読むが `run_diff` は `gate.violations` にしか書かない（違反が success クラスになる）


rizumita:
## Summary

`fslc_rust::outcome::outcome_class` is the declared owner of the "result class ↔ process exit"
mapping. Its `semantic_diff` arm reads a **top-level `violations`** field. `run_diff` never emits
one — it publishes violations under **`gate.violations`** — so `outcome_class` classifies *every*
violating `fslc diff` run as success-class while the process exits `1`.

The sibling `semantic_diff_batch` arm, three lines below, reads `gate.passed` and is correct. The
asymmetry between the two arms is the defect.

This is the general shape of #901 (verify's result-class ↔ exit mapping having no single owner),
reached from a different command.

## Reproduce

```bash
fslc diff examples/db/unsafe_annotated_drop_with_old_server.fsl \
          examples/db/unsafe_annotated_drop_with_old_server.fsl --depth 1
```

Run twice, sequentially; both runs identical.

### Produced

```json
{"result":"semantic_diff","gate":{"forbidden":[],"violations":["impl_violated"],"passed":false}}
```

- Process exit: **1**
- A top-level `violations` key is **absent** from the envelope (checked directly, not inferred).

`outcome_class` applied to that envelope returns the **success** class, because
`output.get("violations")` is `None` and the arm is `is_none_or(Vec::is_empty)`.

### Contrast (same command shape, clean gate)

```bash
fslc diff specs/cart_v1.fsl specs/cart_v1.fsl --depth 1
```

→ `{"result":"semantic_diff","gate":{"forbidden":[],"violations":[],"passed":true}}`, exit **0**.

Same command, different input, different behaviour — so this is not a convention of `diff`, it is
specific to the violating case.

## Expected

`outcome_class` and the process exit agree for every command, which is the property
`check_result_and_exit_status_never_contradict` / `verify_result_and_exit_status_never_contradict`
in `rust/fslc/tests/corpus_check_sweep.rs` assert for `check` and `verify`. Those two tests derive
their class from production code (`outcome_class`) rather than a list, precisely so the law cannot
drift — but they only sweep `check` and `verify`, so `diff` is unguarded.

## Sites

- `rust/fslc/src/outcome.rs` — the `"semantic_diff"` arm reads `output.get("violations")`. Its own
[...]

Issue #1041, opened 2026-09-15
check(refinement): inline `implements` を持つ 153 行の spec で `fslc check` が RSS 28-30GB まで無制限に伸びる（v4.4.1 / v4.5.0 とも再現）


rizumita:
## Summary

`fslc check` on a 153-line spec whose body opens with an inline `implements` block grows
without an apparent bound: **28.0-30.0 GB RSS and still climbing** when run uncapped, on a
machine with 64 GB. It is `check`, not `verify`. Deleting only the 31-line `implements`
block from the same file makes the same command finish in 2 s under 1 MB.

This is **not** a v4.5.0 regression: v4.4.1 reproduces it.

## Reproducer

Two files, in the same directory (the `from` path resolves relative to the spec).
Both are pasted in full below; sha256 of the exact bytes measured:

| file | sha256 | lines |
|---|---|---|
| `claims_M_design.fsl` | `4c738fdedfb24dc7ba8defd043dc9f5b815ad9e1de2d139a51b775f3bc926afb` | 153 |
| `claims_M.fsl` | `552cf50737df54e9f28eb7381b3ace10f6235ecf35558ee015010235f9cb66de` | 124 |

```
fslc check claims_M_design.fsl
```

> [!WARNING]
> Run it under an RSS cap. Uncapped runs drove free memory to 64 MB on a 64 GB machine and
> put every other process at risk of the OOM killer.

## Measured (produced)

All rows: same two frozen files, `check`, no other flags. "capped" rows were stopped by an
8 GiB RSS watchdog (`exit 137`), so the number is a floor, not a peak.

| binary | sha256 | input | RSS | wall | exit |
|---|---|---|---|---|---|
| v4.5.0 release asset | `fdfccc01…e488` | `claims_M_design.fsl` | **8.3 GB (capped)** | 139 s | 137 |
| v4.5.0 release asset | `fdfccc01…e488` | same file, `implements` block deleted | **< 1 MB** | 2 s | **0** |
| v4.5.0 release asset | `fdfccc01…e488` | `claims_M.fsl` (the abstract side) | **12 MB** | 2 s | **0** |
| v4.4.1 release asset | `1d0c2f28…07d1` | `claims_M_design.fsl` | **8.3 GB (capped)** | 25 s | 137 |

Uncapped observations of the same command (three independent runs, terminated by hand):

| elapsed | RSS |
|---|---|
| 1 m 37 s | 29.7 GB |
[...]

PR #1052, opened 2026-09-15
feat(examples): add an insurance-claims corpus at S, M, and L tiers

rizumita:
> ## Status: draft, with `WASM` red and the cause not identified
>
> This pull request is deliberately left as a draft. The `WASM` check is red and **nobody has
> established why.** It is not called a flake here, and it is not called a consequence of this
> corpus, because neither has been shown.
>
> **Measured.** The failing runs of that job are *shorter* than the passing ones — 209 s, 120 s and
> 257 s against 269–384 s — so they are aborting on a thrown exception rather than running out of
> time. This branch also has a **passing** run of the same job at **349 s**, against `main`'s 360 s
> and 362 s. At job level this corpus adds no visible cost.
>
> **Not established.** The same branch content produced a timeout, then a pass, then a timeout.
> `main` and three other branches passed the same job eight times on the same day without one. That
> pattern is consistent with a flaky probe and equally consistent with something specific here; the
> measurement does not separate them.
>
> **So: it has not been shown that the red belongs to this pull request's content.** The first
> failure on this branch did belong to it — a `//` comment containing `from "` was read as an import
> by `rust/fsl-wasm/test-browser.mjs:192`, which is fixed in `7fdb8e88` and filed as #1056 — and the
> `ENOENT` it produced has not appeared since. The current failure is a different one, a
> `browser probe timed out`, and it is unattributed.
>
> The full comparison, the local run under load, and the per-file timings are in
> [this comment](https://github.com/ymm-oss/fsl/pull/1052#issuecomment-5680525609). The harness's
> fixed 90-second budget over a corpus that grows is filed separately as #1055; changing it was
> deliberately not slipped into this pull request.

## Problem

`examples/` had no multi-layer worked example that exercised inline `implements`, an external
refinement mapping, a saga, and a `dbsystem` layer together, and nothing showing how a
specification's size and its verification cost move with each other. Reviewing a change to
refinement or mutation meant reaching for `specs/` fixtures that are each built for one narrow
property.

## What this adds

An insurance-claims corpus at three tiers under `examples/claims/`, plus the one registration its
new refinement mappings need.

| Tier | Files | Shape |
|---|---|---|
| S | `claims_S.fsl` + `negative/pay_before_assessment.fsl` | one `requirements` layer, one negative control |
| M | `claims_M.fsl`, `claims_M_design.fsl`, `claims_ledger_db.fsl` + negative | requirements + design (inline `implements`) + `dbsystem` |
| L | `claims_L_{business,requirements,design,saga}.fsl`, two `*_refines_*.fsl` mappings, `fsl-project.toml` + negative | four-layer chain, external mappings, a `domain` saga |

`rust/fslc/tests/refine_corpus_parity.rs` gains the L tier's two mappings.
`rust/fslc/tests/corpus_check_sweep.rs` structurally excludes the refinement dialect and names that
file as the owner of those cases, so without the rows the two mappings would sit in the tree with
nothing executing them.

**No `fslc` behaviour changes.** Every file is new; the diff is `+2300/-0` across 20 files.

## Contract change

None.

### Run by the author, output read

| Check | Result |
|---|---|
| `fslc check` on **all 13** corpus `.fsl` files | 8 exit 0; 3 negative fixtures exit 2 with their declared `kind`; 2 refinement mappings exit 2 (not `check`-able alone, as the README states) |
| Peak RSS of those 13 `check` runs | max **130,138,112 B (124 MB)**, `claims_M_design.fsl`; second **127,778,816 B (122 MB)**, `claims_L_requirements.fsl`; the other eleven under 8.4 MB. One process at a time under an 8 GiB RSS watchdog, `kern.memorystatus_vm_pressure_level` = 1 before and after every run |
| `fslc mutate` on every mutable spec, uncapped (`--max-mutants 2000`; every run `total < 2000` **and** no `mutant cap … dropped` line) | S 0.9033 · M 0.8889 · M design 0.8240 · L requirements 0.8070 · L design 0.4487 · L business 0.2074 · L saga 0.0182 |
| The same for the three `examples/domain/` files already in the tree | 0.0133 · 0.0175 · 0.0833 — the measurement that scopes the hollow-spec convention below |
| Mutation calibration of `AC-5` | deleting it flips exactly 3 mutants `killed` → `survived` (0.9033 → 0.8943); restored, proven by an empty `git diff` and a matching `sha256` |
| Repair of the six masked oracles, S and M | before: `+Paid`/`+Posted` → `check` exit 0 `ok`, `verify --depth 6` exit 0 `verified`, `verify --depth 8` exit 0 `verified`. After: unchanged at `check` and depth 6, **exit 1 `violated`** at depth 8. `+Approved` → exit 2 `forbidden` `FB-8` throughout, as the accepting control |
| `FB-13`'s mechanism | deleting `LedgerConsistent` turns its silent pass into `kind: "forbidden"`, `id: "FB-13"` — the rejection was the invariant's, not the guard's |
| Duplicate scenario bodies across all six spec files | 0, comparing normalized step sequences. This finds identical bodies only; two different bodies making the same claim are outside it |
| `rust/fslc/tests/corpus_check_sweep.rs` | 5 passed |
| `rust/fslc/tests/corpus_expectation_manifest.rs` | 6 passed |
| `rust/fslc/tests/refine_corpus_parity.rs` | 4 passed — see the note below; it failed first, for a real reason |
| `site reference freshness` (the three pytest modules the workflow runs) | 35 passed |
| `tools/aggregate_changelog.sh check` | PASS, calibrated — an invalid fragment name makes it exit 1 |

### Reported by collaborators, not independently reproduced by the author

- `fslc chain` over `L/fsl-project.toml`.
- The `M` and `L` negative fixtures' `kind` values (their exit codes were measured above).
- The mutation calibration of `AC-6` through `AC-10`; only `AC-5` was re-run independently.
- The per-mutant classification of the 32 S-tier and 149 L-business survivors.

This split is stated because the author orchestrated the work rather than performing all of it. A
line above that is not in the first table has not been observed by the author.

## `refine_corpus_parity` earned its label here

The second commit registers two mappings in that test's `CASES`. Registering them without
regenerating `rust/fslc/tests/goldens/refine_corpus_projection.v1.json` makes the test fail:

```
refine_corpus_parity.rs:970
assertion `left == right` failed: the projection golden and live CASES must have exactly
the same rows; regenerate only after reviewing the observed envelope diff
```

That is what happened on the first run of this branch, and it is worth stating plainly: the test is
a **detector**, not a preservation control, and this pull request is the execution that proves the
label. The golden was then regenerated with `UPDATE_FSL_REFINEMENT_GOLDEN=1`, its diff reviewed, and
the test re-run.

The first run of that test against this change was also green for the wrong reason — it was run in a
worktree that did not contain the change. The gate worked correctly and reported correctly; the
green was attached to a different tree. Every gate result in the table above was taken from a run
whose own output names `.../worktrees/fsl/claims-corpus/...`.

## Six of this corpus's own oracles could not fail, and that is why the corpus is worth reading

`fslc mutate` reported 0.9033 on `claims_S.fsl`. Inside that number, six `forbidden` scenarios that
read as oracles for one property — that `lapse` is excluded from the post-payment stages — could not
fail. Widening the guard by one enum member, with its `days >= LAPSE_AFTER` clause untouched, left
`check` and `verify --depth 6` green for `Paid` and `Posted`. Only `FB-8` was a detector.

Three different mechanisms were behind the one symptom, and each needed a different repair:

| scenario | why it could not fail | repair |
|---|---|---|
| `FB-5`, `FB-11`, `FB-12` | reached `lapse` with `days == 0`, so the day guard rejected the call before the stage clause was read | raise `days` first, as `FB-8` already did |
| `FB-13`, `FB-14` | did raise `days`, and still passed — **measured** by deleting `LedgerConsistent` and re-running: the rejection came from the ledger invariant, not the stage guard | moved to `trans LapseNotFromPaid` / `LapseNotFromPosted`, because `expect rejected` cannot separate the two |
| `FB-9` | its conjunct `claim[c].st != Lapsed` in `advance_day` is unreachable (`Lapsed` implies `days == LAPSE_AFTER`) | deleted; widening `Day` does not revive it, since `LapsedMeetsWaitingPeriod` ties the two independently of the bound |

Deleting `FB-5`'s mask made it byte-identical to `FB-8`, so it is gone: one property was being counted
under two ids.

### Three things in this pull request are correct and unguarded

Stated together, because each on its own reads like a smaller admission than it is. All three are filed as #1055, which records the gap rather than adding a check in a hurry.

1. **The two `trans` properties need `verify --depth 8`.** `check` and `verify --depth 6` produce
   byte-identical output for `Paid`/`Posted` before and after the repair. `examples/claims/README.md`
   states it in these words, and they are the words that matter here too:

   > **`rust/fslc/tests/corpus_check_sweep.rs`, the CI job that walks this corpus, only runs `check`
   > (line ~102) and `verify --depth 2 --deadlock ignore --engine bmc` (lines ~128-134) — depth 2,
   > not 8.** No CI gate in this repository currently exercises
   > `LapseNotFromPaid`/`LapseNotFromPosted`. This is not a claim that the fix is wrong: the property
   > is correctly stated and `verify --depth 8` (below) does detect a violation of it. It is a claim
   > about what *is* and *is not* covered by an automated gate today, stated because the whole point
   > of this pass was not repeating exactly this gap.
2. **`claims_L_saga.fsl` declares no property at all.** That is consistent with the `domain`
   dialect's three existing examples, and the hollow-spec convention does not apply there (see
   below), but nothing in this file can fail.
3. **The reasons recorded at each edit site are comments.** Deleting them breaks no gate. They raise
   the chance the next person does not re-add the conjunct this change removed; they do not ensure it.

### The other result worth reading before trusting a kill rate

The "kill rate below ~10% means a hollow spec" convention does not apply to the `domain` dialect. The
three `domain` examples already merged measure 0.0133, 0.0175, and 0.0833 — all below it. The saga
added here measures 0.0182, above the nearest precedent. `mutate` gained `Domain` support only in
#727. The README states that scope with those four measurements, and no oracle was added to the saga
to move the number.

## Premises this rests on that have not been verified

- That the two 124 MB files are the two carrying inline `implements` is **consistent with** #1041's
  account of where the cost comes from. With n = 2 it is not evidence for it.
- The survivor classifications are read from the specs and from `mutate`'s own `survived` verdict.
  None was confirmed by adding an oracle and watching the mutant flip to `killed`.
- `fslc mutate` refuses `claims_ledger_db.fsl` with `result: "error"`, `kind: "semantics"`,
  `"mutate expects a spec-like FSL file"` (`rust/fslc/src/main.rs:10925-10930`). The README records
  this as current behaviour; `docs/DESIGN-mutate.md` does not mention the `dbsystem` dialect either
  way. No issue was filed: the refusal is fail-closed and diagnosed, which is not the case
  `AGENTS.md` treats as a soundness defect.
- The deletion calibration of `AC-6` through `AC-10` is a collaborator's measurement. Only `AC-5` was
  re-run independently by the author.
- `fslc chain` over `L/fsl-project.toml` was not run by the author.
- `product gate` and the three `native Z3 4.16` lanes are skipped on pull requests; the post-merge
  `main` run is what exercises them. A green check list here is 18 green, 4 skipped, not 22 green.
- The three unguarded items in the section above are the same shape as this list, not a milder one:
  a property that exists, a file that declares none, and a comment that no gate protects.

## Not in this pull request

The original plan included cardinality variants of the L tier as a scaling axis. `check` on the
larger members exceeds 8 GiB — see #1041 — which puts them outside every population that must
`check` cleanly (`specs/`, `examples/`, `rust/fslc/tests/fixtures/`). The variants and their
generator live in #1041 instead, and `examples/claims/L/README.md` records that substitution with
the generator's `sha256`. #1041 and #1043 are both open and unfixed; nothing here works around
either.

⚠️ #1041's reproducer uses the filenames `claims_M.fsl` and `claims_M_design.fsl`. Those are
different files from the ones added here — different `sha256`, different line counts, and the
corpus versions `check` in 124 MB and 8 MB. A comment on #1041 records the comparison.

## Review ledger

The obligation ledger for this work is `.claude/work/REVIEW-MEM-EX-OBLIGATIONS.md` in the authoring
worktree (not committed — it is working evidence, not repository authority, per `AGENTS.md`'s
knowledge-distillation rule). Its counts: **8 obligations, 8 completed, 8 evidence-supported, 5
verified** (verified meaning the author executed it and read the output).

It is recorded there, and repeated here, that the ledger did not find the largest defect in this
pull request. All six masked oracles fell outside its eight obligations, because those obligations
were written about what this change *added* and not about the existing controls the additions leaned
on. The ledger is useful as a record of which questions were answered and with what; it is not
evidence that the pull request was reviewed.


rizumita:
The `WASM` check on this branch is red, and the measurement does not support the explanation I
started with. Recording it here so the next person does not repeat it.

## What I thought, and what CI says

The first failure was real and is fixed: `rust/fsl-wasm/test-browser.mjs:192` matches
`/\b(?:from|refinement)\s+"([^"]+)"/g` over raw source with no comment handling, so a `//` comment
containing `from "an unrelated` was read as an import (`ENOENT` on a path with a newline in it).
Reworded in `7fdb8e88`, filed as #1056. `ENOENT` appears 0 times in the log since.

The second failure is `browser probe timed out` at `test-browser.mjs:442`. My first reading was that
this corpus added ~26 parity cases (one `check` and one `verify` per spec) and pushed the probe past
its budget — `for (let attempt = 0; attempt < 360; …) await delay(250)`, i.e. 90 seconds.

**Job durations say otherwise.** Same workflow, same job name, today:

| conclusion | branch | seconds |
|---|---|---:|
| failure | `examples/claims-corpus` | 209 |
| failure | `examples/claims-corpus` | 120 |
| success | `main` | 360 |
| success | `fix/1051-check-scope-docs` | 384 |
| success | `main` | 362 |
| **success** | **`examples/claims-corpus`** | **349** |
| failure | `examples/claims-corpus` | 257 |
| success | `fix/1051`, `fix/1044`, `main`, `fix/1043` | 269–377 |

The failing runs are **shorter** than the passing ones — they abort on the thrown exception rather
than running long. And this branch's one passing run, 349 s, sits with `main`'s 360 and 362. At job
level this corpus adds no visible cost.

## What is established, and what is not

- **Established:** the corpus does not increase the job's duration. The first failure's cause is
  found, fixed here, and filed as #1056.
- **Not established:** that the timeout is caused by this branch. The same content produced a
  timeout, then a pass, then a timeout. That pattern is consistent with a flaky probe — this
  repository already has a note that the browser leg can hang — but `main` and three other branches
  passed eight times today without one, so "flake" is not established either.

I am not changing `test-browser.mjs` in this pull request. The 90-second budget is a fixed deadline
over a corpus that grows, which is worth its own issue and its own four-conditions argument before a
check is written; that is #1055's territory, not a fix to slip in here.

A local run under load also timed out (wall 998 s, load 8.25 → 13.17 across the run). That is a
data point, not a verdict: under load a timeout cannot separate "this corpus is expensive" from "this
machine is busy".

## Per-file cost, since it came up

Native `fslc` on this corpus, two replicates, taken with no other `fslc` running:

| file | rep 1 `check` | rep 2 `check` |
|---|---:|---:|
| `M/claims_M_design.fsl` | 2.84 s | 3.60 s |
| `L/claims_L_requirements.fsl` | 1.68 s | 2.56 s |
| the other eleven | 0.09–0.17 s | 0.11–0.33 s |

The replicate spread is large (+27 % and +52 %), so the absolute numbers are not usable. The ratio is
not: the two files carrying an inline `implements` block cost ten to thirty times the rest, which is
well outside that spread. Those are the same two files that lead the `check` RSS table in the
description. This is measured with `check`, whose refinement depth is fixed at 8; it says nothing
about what a user-chosen `verify --depth N` would cost.

Issue #1082, opened 2026-09-18
[Draft] roadmap: install the Agent Skills from the binary, and make mise a first-class route


mpyw:
## Status

Draft. This is a roadmap, not an accepted design. Several parts are unverified.
Each phase below should become its own issue once the spike ahead of it lands.
Sections marked **Unverified** have no evidence yet.

## Problem

An Agent Skill only helps when it is installed. Today, whether a user gets the
skills depends on which install route they picked.

| Gap | Evidence |
| --- | --- |
| The single-executable download gets no skills at all | `README.md:84-88`, "no skill integration" |
| Windows ships `fslc-windows-x64.exe`, but `install.sh` is bash only | `install.sh` has no PowerShell path |
| Uninstall is a hand-written `rm -rf` list | `README.md:68` names ten paths |
| Copying skills into another project leaves no version record | `README.md:302` tells users to `cp -R` |
| A mise user must hand-write a `[tool_alias]` block | mise never reads a tool repository's own `mise.toml` |

The skill references are coupled to the binary by contract. `AGENTS.md` states
that a language feature moves with the FSL skill references. Nothing enforces
that coupling at install time.

## What is already true

Measured on `main` at `294b4897`. Recorded so it is not re-litigated.

| Fact | Value |
| --- | --- |
| Shipped binaries | `fslc` and `fslc-lsp` only |
| Other `src/bin/` targets | 8, none of them released |
| Skill payload | 22 files, 312K, about 100K gzipped |
| `fslc-macos-arm64` | 36,178,928 bytes |
| `fslc-lsp-macos-arm64` | 5,676,528 bytes |
| `install.sh` skill handling | already a symlink into the versioned payload (`install.sh:279-291`) |

`install.sh` already isolates the payload under `~/.local/share/fsl`. It links
`~/.claude/skills/fsl*` at `current/skills/*`. That link is why the install
script route has no version drift. Only the other routes do.

### mise, measured with 2026.8.4

| Command | Result |
| --- | --- |
| `mise install github:ymm-oss/fsl@4.6.0` | installs `fslc`, verifies attestations and SLSA provenance |
| Same spec twice with different `matching` | second install is a no-op, both keys resolve to one directory |
| `[tool_alias]` with two aliases | **works**, `installs/fslc/4.6.0/fslc` and `installs/fslc-lsp/4.6.0/fslc-lsp` |

> [!IMPORTANT]
[...]

PR #1083, opened 2026-09-18
docs(readme): document installing fslc and fslc-lsp with mise

mpyw:
## Problem

mise installs `fslc` today, but the README never mentions it. Two things are not
discoverable from the release page.

- One repository publishes two commands. `[tools]` keyed by the same backend
  resolves to one install directory, so the second entry is a no-op. Each command
  needs its own `[tool_alias]`.
- `matching` is a case-sensitive substring test. `matching = "fslc-"` also matches
  `fslc-lsp-macos-arm64`, so it cannot select `fslc`.

Without both facts, a reader either gets one command or the wrong one.

## Contract change

None. Documentation only. No release asset, installer, or code change.

## Evidence

Measured with mise 2026.8.4 against the published v4.6.0 release, in a scratch
`MISE_DATA_DIR`.

| Configuration | Result |
| --- | --- |
| `github:ymm-oss/fsl@4.6.0` alone | installs `fslc` only |
| The same backend twice, differing only by `matching` | second install is a no-op, one directory `github-ymm-oss-fsl/4.6.0` |
| The `[tool_alias]` pair in this PR | `installs/fslc/4.6.0/fslc` and `installs/fslc-lsp/4.6.0/fslc-lsp` |

`fslc version` from the first path reports `fslc 4.6.0`. The second is the LSP
server, which exits with a protocol error when run without a client, as expected.
mise reported checksum, GitHub artifact attestation, and SLSA provenance
verification on every install.

`tools/aggregate_changelog.sh check` passes.

## Linked issue

First step of #1082.

## Documentation and skill impact

`README.md` gains a `### With mise` section and an entry in the install-route
list. No skill file changes. The new section states that this route installs the
two commands only, not the Agent Skills. Later steps of #1082 change that.

## Note on the stack

This is the first of a stacked series. The branch is on a fork, so every pull
request in the series targets `main` here. Until this one merges, later pull
requests will show these commits in their diff as well.

---


rizumita (changes requested):
Thanks for writing this up — the `[tool_alias]` pairing and the `matching` substring trap are exactly the kind of thing nobody finds from the release page, and it's great to have them measured.

A couple of things before this can go in:

1. **Attestation / SLSA claim.** The README says mise verifies "the GitHub artifact attestations, and the SLSA provenance". As far as I can tell our releases don't publish either: `release.yml` has no `actions/attest-build-provenance` step (and no `attestations: write` / `id-token: write` permission), and `GET /repos/ymm-oss/fsl/attestations/<digest>` returns 404 for the v4.6.0 `fslc-*`, `fslc-lsp-*` and `fsl-skills.tar.gz` digests (the same query against `cli/cli` returns attestations, so it's not the API). Could you drop that part of the sentence, or limit it to the checksum? Adding real provenance to the release workflow would be welcome, but as a separate PR.
2. **Pinned `4.6.0`.** We removed a hard-coded "current release" from the README in #827 because nothing in the release procedure updates it. Could the snippet use `version = "latest"` (or an obvious placeholder) with a short note to pin a version for reproducibility and to keep `fslc` and `fslc-lsp` on the same version?

Nits (take or leave): `matching_regex = "^fslc-lsp-"` would mirror the `fslc` entry; and mentioning the minimum mise version that supports backend aliases in `[tool_alias]` would help people on older installs.

CI hasn't actually run on this branch yet (the fork runs never started); I'll approve the workflows once you push an update.

PR #1084, opened 2026-09-18
feat(skills): ship every skill in skills/ instead of a hardcoded six

mpyw:
## Problem

`skills/` holds eight skills. Six of them reach users.

The same six-name list was written out twice, with nothing comparing the two.

| Location | Role |
| --- | --- |
| `.github/workflows/release.yml:280-286` | what enters `fsl-skills.tar.gz` |
| `install.sh:225` and `install.sh:281` | what gets linked into `~/.claude/skills/` |

Adding a skill to `skills/` therefore shipped it nowhere. `fsl-from-code` and
`fsl-requirements-document` have never been in a release bundle.

## Contract change

The distributed skill set is now derived from the payload, not from a list.

- The workflow bundles every directory under `skills/`. `skills/README.md` is a
  file, so it is excluded by construction rather than by name.
- The installer links every directory it finds in the bundle.
- `fsl-from-code` and `fsl-requirements-document` ship for the first time.
- The installer fails when the payload holds no skill at all, instead of
  silently linking nothing.

The bundle step now also asserts that `skills/fsl/SKILL.md` is present before
the checksum is taken. A silently empty bundle would otherwise pass.

## Evidence

The bundle pipeline, run against the working tree:

| Check | Result |
| --- | --- |
| Top-level entries in the archive | the eight skill directories |
| `skills/README.md` present | no |
| Guard `skills/fsl/SKILL.md` | passes |

`payload_skill_names` against an extracted bundle with a `README.md` added
beside the directories returns eight names, sorted, and does not return
`README.md`. Against a payload holding no directory it returns nothing, so the
new guard fires.

`bash -n install.sh` passes. `shellcheck -S warning install.sh` is clean.
`tools/aggregate_changelog.sh check` passes.

Packing was exercised with bsdtar locally. The workflow runs GNU tar on
ubuntu-24.04, where `--null --files-from -` and `sort -z` are native.

## Observed, not changed here

`install.sh:129` keeps its own copy of the six-name list, inside the v3.0.0
compatibility path. That branch is unreachable. `RELEASE_TAG` always comes from
the latest-release API (`install.sh:58-67`) with no override, so the
`!= "v3.0.0"` guard above it is always true.

Making that branch generic would also change what it installs. v3.0.0's
`skills/` holds seven directories, including `fsl-from-code`. Deleting the
branch outright is the better change, and it belongs in its own pull request.

## Linked issue

Second step of #1082. Implements the decision recorded there under
"Which skills ship".

## Documentation and skill impact

`README.md` drops the sentence naming the installed subset, and the uninstall
command gains the two new paths. No skill file content changes.

## Relationship to #1083

Independent. Both pull requests edit `README.md`, but in regions that do not
overlap. #1083 changes the install-route list and adds a section. This one
changes the uninstall command and the paragraph naming the installed subset.
They merge in either order without a conflict, verified locally.

This branch sits directly on `main`, so the diff here is this change only.

---


rizumita (commented):
Thanks — deriving the shipped set from the payload instead of two hand-kept lists is the right fix, and it matches `AGENTS.md` ("`skills/` holds distributable canonical skills"). I confirmed the v4.6.0 `fsl-skills.tar.gz` holds only the six, and that the new `install.sh` still behaves correctly against that existing 6-skill bundle, so merging before the next release is safe. The empty-bundle guards on both sides are a nice touch.

A few suggestions, none blocking:

1. **Key on `SKILL.md`, not "is a directory".** Excluding `README.md` by construction works because it's a file, but a future non-skill directory under `skills/` (shared assets, a README-only folder, a dot-directory) would be bundled and linked into `~/.claude/skills`. Selecting `skills/*/SKILL.md` in the workflow, and adding `[ -f "$dir/SKILL.md" ] || continue` in `payload_skill_names`, would make both sides agree on what a skill is (today `find -type d` also picks up hidden directories that the `*/` glob in the installer skips).
2. **README timing.** "The install script above links every skill above" only becomes true once the next release is cut; until then `install.sh` still gets the six from v4.6.0. Maybe "starting with the next release"?
3. **Users who followed the old advice.** The previous README told people to copy `fsl-from-code` / `fsl-requirements-document` into `~/.claude/skills` by hand; on the next install those copies will be moved to `*.pre-native-v3`. Nothing is lost, but a short mention in the changelog fragment would avoid surprise.

Nits: the uninstall `rm -rf` line is effectively still a hand-kept list — removing only links that point into `~/.local/share/fsl/current/skills/` would avoid that; and the last line of the rewritten README paragraph is wider than the surrounding wrap.

I've approved the workflow runs; I'll approve the PR once CI is green.

rizumita (approved):
CI is green on 19ddb42. Approving — the suggestions in my earlier review are non-blocking and can land as follow-ups.

PR #1085, opened 2026-09-18
feat(skills): carry the Agent Skills in the binary and add fslc skills install

mpyw:
Resolves the Phase 2 and Phase 3 steps of #1082.

`fslc` now carries the Agent Skills inside the binary and places them itself.

## Why

Only one install route shipped the skills, and it shipped them separately.

| Route | Before | After |
| --- | --- | --- |
| `install.sh` | linked, as a separate download | `fslc skills install --user` |
| Single executable | **nothing** | `fslc skills install` |
| Windows | **nothing** | `fslc skills install` |
| mise | **nothing** | `fslc skills install` |

Nothing tied the shipped skills to the `fslc` that reads them. `AGENTS.md`
requires a language feature to move with its skill references, so the two
drifting apart is a correctness problem, not a packaging one.

`build.rs` walks `skills/` and generates an `include_str!` table. The skills
and the binary that reads them are now one artifact.

## The command

```
fslc skills install   [--dir <path>] [--user] [--force] [--dry-run]
fslc skills status    [--dir <path>] [--user]
fslc skills uninstall [--dir <path>] [--user] [--force] [--dry-run]
```

Stdout is the usual JSON envelope.

```json
{
  "fsl": "1.0",
  "command": "skills install",
  "scope": "project",
  "directory": "/tmp/x/s",
  "fslc_version": "4.6.0",
  "dry_run": false,
  "result": "installed",
  "written": 21,
  "skills": ["fsl", "fsl-business", "..."]
}
```

The exit code is derived from `result`, so a response cannot report success
and exit non-zero.

| `result` | Exit |
| --- | --- |
| `installed`, `up_to_date`, `removed` | 0 |
| `blocked`, `drifted`, `stale`, `not_installed` | 1 |

`status` exits non-zero when the installed set is stale or drifted. It works
as a check, not only as a report.

## Two scopes

| Scope | Where | How | Why that way |
| --- | --- | --- | --- |
| **project**, the default | `.claude/skills` in the repository | copies | they stay put when another project selects a different `fslc` |
| **user**, `--user` | `~/.claude/skills` | links through `current` | an upgrade is one repointed pointer |

The user scope reproduces the shape `install.sh` already builds:

```
~/.claude/skills/fsl -> <data>/current/skills/fsl
<data>/current       -> <data>/releases/v4.6.0-25984eeca29e
```

> [!NOTE]
> The vocabulary is **project** and **user**, matching `gh skill install`.
> `~/.local/share` and `~/.local/bin` keep their names, because those are
> XDG's, not ours.

### Writing the user-wide directory by accident

The project search starts at the working directory and stops at the
repository root. The first existing `.claude` wins. Without one, the
repository root gets a new one.

> [!IMPORTANT]
> The bound is the point. `~/.claude` exists for every Claude Code user. An
> unbounded walk from anywhere under `$HOME` always reaches it, and quietly
> turns a project install into a machine-wide one.

Both refusals, run against the built binary:

```console
$ cd $HOME/project && fslc skills install      # no .claude, no repository
"no .claude directory at /.../project/.claude and no repository root to
 create one in; pass --dir, or use --user"                            exit 2

$ cd $HOME && fslc skills install              # the home directory itself
"/.../.claude is the user-wide directory; use --user to install there" exit 2
```

`project::location::resolve_dir` takes an `exists` predicate instead of
calling the filesystem, so both refusals are unit tested without a real home.

> [!CAUTION]
> Review found this guard escapable, and the fix is in this branch.
> `current_dir` resolves symbolic links and `$HOME` does not. With a
> symlinked home the walk landed on the real path, the guard compared it
> against the link, and the install wrote the user-wide directory while
> reporting `"scope": "project"`. Measured before the fix: exit 0, `21`
> files written. After: exit 2, nothing written.

### Leaving the directory it was given

A manifest key is checked as text. Text is not enough, and neither is a
spelling: what matters is the place a path resolves to.

> [!CAUTION]
> Review found this escapable, and the fix is in this branch. A symbolic link
> at any **parent** component made a contained-looking key resolve elsewhere,
> and `create_dir_all`, `write` and `remove_file` all followed it. A
> repository can carry both halves, so one `fslc skills install` in a fresh
> clone wrote outside the directory it named, and a manifest naming a file
> under the link made `uninstall` delete outside it.

Measured, with no `--force` anywhere:

| | Before | After |
| --- | --- | --- |
| `install --dir <skills>` with `<skills>/fsl -> victim/` | exit 0, `"installed"`, 10 files in `victim/` | exit 2, nothing written |
| `uninstall` with a manifest naming `fsl/key` | exit 0, `"removed"`, the file outside is gone | exit 2, the file survives |

Every component is now checked before a write or a removal. `--force` does not
reach it, because a directory somebody linked elsewhere was never this
installation's to write.

### Overwriting something it did not write

A manifest records every path written and the digest written there.

| State of the path | `install` | `uninstall` |
| --- | --- | --- |
| absent | write | skip |
| **managed**, ours and unchanged | write if stale, else keep | remove |
| **modified**, ours and hand edited | blocked, `--force` overrides | blocked, `--force` overrides |
| **foreign**, a link no manifest of ours records | blocked, `--force` overrides | **skip** |
| **occupied**, a real directory or file, user scope | blocked, `--force` overrides | **skip** |

`--force` reaching a foreign entry is deliberate. A user who wrote a skill by
hand, or installed one with `gh skill install`, may now want ours. Their only
other route is deleting the directory by hand.

> [!WARNING]
> `uninstall` keeps skipping a foreign entry even under `--force`. Removing
> it would delete something `fslc` never wrote.

`--force` differs in kind between the scopes, so the states do too. A project
install replaces the files it is about to write. A user install replaces
whatever sits at `~/.claude/skills/<skill>`, and a real directory there goes
with everything under it. That path reports `occupied`, not `foreign`, so the
two are distinguishable before anyone reaches for `--force`.

A symbolic link where a skill file goes is never written through. Under
`--force` the link is removed and our file is written in its place. What the
link aimed at is not ours to touch.

### Leaving a skill behind on an upgrade

A binary that carries fewer skills than the one before it has to take the
difference out. The manifest is what makes that possible: it records what the
previous run placed, so the current run can tell a retirement from a file
somebody else wrote.

| Scope | A skill this binary no longer carries |
| --- | --- |
| project | its files are removed, and a directory they emptied is pruned |
| user | its link is removed, and `status` reports `drifted` until it is |

> [!CAUTION]
> The user scope did not do this, and the fix is in this branch. The link
> survived, and the manifest that named it was rewritten without it, so no
> later `uninstall` could take it back. Once `current` moved, it pointed at
> nothing. Measured before the fix: `"up_to_date"`, exit 0, a dangling
> `~/.claude/skills/fsl-retired` surviving both `install` and `uninstall`.

### Stranding the commands

`install.sh` links `~/.local/bin/fslc` at `current/bin/fslc`. Moving `current`
to a release with no `bin/` would leave the command unreachable. That is
refused before any write, `--force` included.

## Why a separate crate

The skills module started inside the `fslc-rust` library. `fsl-lsp` and
`fsl-wasm` link that library with default features off. The module needs
`sha2`, which `fslc-rust` declares optional under `native-cli`.

| Command | Module inside `fslc-rust` | After the split |
| --- | --- | --- |
| `cargo check -p fslc-rust --no-default-features --lib` | `error[E0432]: unresolved import sha2` | passes |
| `cargo check -p fsl-wasm` | the same, through the library | passes |
| `cargo check -p fsl-lsp` | the same, through the library | passes |
| `cargo build --workspace` | **passes** | passes |

> [!IMPORTANT]
> The workspace build passed the whole time. Feature unification turns
> `native-cli` on for the binary, and every other crate then sees `sha2`. No
> existing gate can observe this class of break.

Making `sha2` unconditional would have hidden it. The Worker and the editor
would still compile every embedded file, into builds that can never install
one.

`tools/check-native-integration.sh` now checks the boundary in both
directions.

| Assertion | Meaning |
| --- | --- |
| no `fsl-skills` in the `fsl-wasm` or `fsl-lsp` tree | the payload cannot reach them again |
| no FSL stack crate in the `fsl-skills` tree | the crate stays outside the language |

### The design record

`docs/DESIGN-rust-components.md` froze the crate count at eleven. Its own
section 10 lists the trigger that fired here:

> LSP or Worker cannot reuse a shared diagnostic/result without importing
> inappropriate native dependencies.

Section 11 records the amendment, and why this is not the C2 split that the
same document rejects.

| C2 rejection ground | Applies here? |
| --- | --- |
| a crate without consumer need | No. The boundary **is** the need |
| one workspace version, no independent lifecycle | Unchanged, and shared |
| moved code duplicates composition policy | No. It composes nothing |

## Shape

```
rust/fsl-skills/
├── build.rs              generates the include_str! table
└── src/
    ├── lib.rs            docs, shared names, re-exports
    ├── embedded.rs       which skills the binary carries
    ├── installation.rs   which of the two scopes a run means
    ├── installation/
    │   ├── manifest.rs   what a previous run recorded
    │   ├── plan.rs       what a run would do, before it does any of it
    │   ├── project.rs    a project's copies
    │   │   └── location.rs   the bounded search
    │   └── user.rs       the machine's links
    │       ├── location.rs   the data directory and the release name
    │       ├── payload.rs    the release payload and `current`
    │       └── symlink.rs    link placement, and the Windows limit
    └── cli/
        ├── command.rs    arguments and subcommands
        ├── report.rs     the JSON envelope
        └── selection.rs  flags to installation
```

Each outcome is a struct, so the JSON contract is the type. `Listing` is an
enum, so the variant name supplies the key.

```rust
struct ProjectInstall<'a> {
    #[serde(flatten)] header: Header,
    result: Verdict,
    written: usize,
    skills: Vec<&'static str>,
    #[serde(flatten)] listing: Listing<'a>,
}
```

Visibility was re-audited after the split, by compiler rather than by eye.

| Question | Method | Result |
| --- | --- | --- |
| Is any `pub` unreachable from outside? | `-W unreachable_pub` | 0 |
| Can any `pub(crate)` be narrower? | narrow it, then compile | 5 narrowed, 4 proved necessary |

| | `pub` | `pub(crate)` | `pub(super)` |
| --- | --- | --- | --- |
| Items | 46 | 22 | 32 |

`fslc` names exactly one of the 46: `fsl_skills::cli::run`.

## Cost

| Measured on the release build | Value |
| --- | --- |
| Skills | 8 |
| Embedded files | 21 |
| Payload | 257,751 bytes |
| Binary | 37,369,568 bytes |
| Payload share | **0.69%** |

## What adversarial review turned up

Two read-only reviewers went over the branch. Every finding below was
reproduced against the built binary before it was fixed.

| What was wrong | Was | Now |
| --- | --- | --- |
| a symlinked parent component carried writes and deletes outside the directory | exit 0, files written and deleted outside | refused, `--force` included |
| a symlinked `$HOME` slipped past the user-scope refusal | exit 0, `"scope": "project"` into `~/.claude` | refused |
| a project install that failed partway recorded nothing | its own writes read as foreign, `uninstall` said `not_installed` | records what it placed, then reports the failure |
| `uninstall --user` removed this binary's payload, not the recorded one | the recorded payload stayed, unreferenced and unremovable | removes the release the manifest names |
| `--dir` skipped the user-scope refusal | a project manifest in `~/.claude/skills` wedged `--user` for good | refused, and `--dir` elsewhere is unchanged |
| a dry run reported `payload_written` as a flag | dry run `1`, real run `21` | both report the count that differs |
| Windows wrote the whole payload, then failed at the link | one orphan per attempt | refuses before writing a byte |
| Windows sets `USERPROFILE`, and only `HOME` was read | a project install failed with "HOME is not set" | `USERPROFILE` is the fallback |
| a directory where a skill file goes reported "failed to read" | `EISDIR` verbatim | names the cause |
| `install.sh` treated a refusal as a failure | aborted after the binaries were live | warns, and only a real failure is fatal |
| a user-scope upgrade left a retired skill linked | `"up_to_date"`, a dangling link nothing could remove | the link is removed, and `status` sees it until then |
| the containment walk never looked at the skills directory itself | a repository linking `.claude/skills` at `~/.claude/skills` wrote there, reporting project scope | the guard compares the place a path resolves to |
| `manifest.release` was unvalidated | a recorded absolute path made `uninstall --user` recursively remove it | a release is one component, checked on read |
| a link key could be `a/b` | it reached through a symlink inside the skills directory | a link key is one component, checked on read |
| recording a partial write replaced the previous manifest | 21 entries became 0 with 21 files on disk, and `uninstall` then removed none of them | what the run did not reach stays recorded |
| `--dir` compared spellings | `~/.claude/x/../skills` got past the refusal | `.` and `..` are folded before the comparison |
| the manifest's staging file followed a symbolic link | `.fslc-skills.json.<pid>` could be planted, so the manifest landed outside the directory and the rename left a link in its place | the name is unguessable and `create_new` refuses what is already there |
| `--dir` took the next argument whatever it was | `--dir --force` made a directory called `--force` and dropped the flag | an option where a path belongs is refused, as the contract already declared |

Several tests asserted less than their names claimed, and were tightened. Two
wrapped every assertion in `if let Err(...)` to skip the root case, so they
also passed on the exact regression they guarded. They now measure whether the
setup took effect and assert unconditionally when it did. The
uninstall test now checks the payload is gone, which is what let the orphan
above hide.

> [!IMPORTANT]
> One of these was a regression a previous fix introduced. Recording a partial
> write was meant to stop `uninstall` disowning files, and instead it replaced
> a complete manifest with the prefix one run reached. Its test carries a
> rejecting control: with the fix disabled the manifest holds 0 entries where
> 21 files sit on disk, and the test fails.

The Windows findings are read-only. This branch has no Windows host, so the
fix is reasoned from the source, not measured.

## Evidence

109 test functions. 94 in `fsl-skills`, 15 in `fslc-rust`.

The `fslc-rust` ones run the built binary. Setting an environment variable is
`unsafe` in Rust 2024, so the user-scope tests drive a subprocess.

Every trap has a rejecting control.

| Trap | Control |
| --- | --- |
| a manifest path escaping its directory | refused, nothing written |
| a symbolic link where a skill file goes | the link target stays absent |
| a symlinked `$HOME` | the user-scope refusal still fires |
| a `current` holding the installed commands | refused before any write |
| a manifest recording the other scope | every subcommand refuses |
| a skill this binary no longer embeds, project | its files go, and the directory they emptied is pruned |
| a skill this binary no longer embeds, user | its link goes, in the plan and on disk |
| a planted staging name for the manifest | the manifest stays a file this run wrote |
| an option where `--dir` wants a path | refused, and no directory of that name appears |
| a partial failure | what was placed is still recorded |

| Check | Result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass |
| `cargo build --workspace --locked` | pass |
| `cargo test -p fsl-skills` | 94 pass, 0 fail |
| `cargo test -p fslc-rust --test skills_user_cli` | 15 pass, 0 fail |
| `cargo test -p fslc-rust --test error_envelope_parity --test native_integration --test induction_cli_contract` | 32 pass, 0 fail |
| `tools/aggregate_changelog.sh check` | pass |
| `tools/check-shell-scripts.py` | pass, 10 scripts |
| `pytest` site reference, manual, refresh | 35 pass, 0 fail |

CI names packages explicitly on the macOS, Windows, and production lanes, so
those now list `-p fsl-skills`. Without it the moved tests would have stopped
running on Windows.

## Deliberate choices

| Choice | Reason |
| --- | --- |
| skills stay out of `FSLC_IMPLEMENTATION_FINGERPRINT` | that digest identifies what the verifier proves with, and prose does not change it |
| `skills <verb>`, not `init` and `deinit` | in a specification language CLI, `init` reads as "scaffold a project" |
| nothing is installed implicitly | the command writes into `.claude/` only when asked, and prints the directory it chose |
| the subcommand stays plural | `gh skill install` names one skill. This one always places the whole set |
| no lock file | local IO over small files, and the manifest is already written last and atomically. A killed run used to leave a lock that every later run refused |

## Limits

`--user` needs symbolic links, so it fails on Windows, where creating one
needs privilege. A project install copies, and works there.

## Note on the stack

Built on #1083 and #1084. The branch is on a fork, so this targets `main`
here and its diff shows those commits until they merge.

---

https://claude.ai/code/session_01PngwJbm7WsZHmMXdTKicd1


rust/fslc/src/skills_install.rs
```diff
@@ -0,0 +1,790 @@
...
+use sha2::{Digest, Sha256};
+
+use crate::skills::EMBEDDED_SKILL_FILES;
+
+/// Schema of the manifest written beside the installed skills.
+pub const MANIFEST_SCHEMA: &str = "fslc.skills.v1";
+/// Manifest file name, inside the skills directory.
+pub const MANIFEST_FILE: &str = ".fslc-skills.json";
+/// The agent directory this tool looks for and writes into.
+pub const AGENT_DIR: &str = ".claude";
```
mpyw: TODO: Claude 以外のエージェントにも対応する余地はあるが，現状の README が Claude 決め打ちなのでいったん OK

rust/fslc/src/skills_install.rs
```diff
@@ -0,0 +1,790 @@
...
+    ///
+    /// A project install fills this. A user install writes links, not files.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub entries: BTreeMap<String, String>,
+    /// Links created, as a skill name to the target it was pointed at.
+    ///
+    /// A user install fills this. Recording the target is what lets removal
+    /// tell a link this tool made from one somebody else put there.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub links: BTreeMap<String, String>,
```
mpyw: IMO: どういう対応関係のハッシュマップなのか分かりにくいのでコメントで補足求む

rust/fslc/src/skills_install.rs
```diff
@@ -0,0 +1,790 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum EntryState {
+    /// Nothing is there.
+    Absent,
+    /// Recorded by the manifest and unchanged since it was written.
+    Managed,
+    /// Recorded by the manifest and edited since it was written.
+    Modified,
+    /// Present, and recorded by no manifest of ours.
+    Foreign,
```
mpyw: IMO: Foreign がどういう意味なのか分かりにくいかも？

rust/fslc/src/skills_install.rs
```diff
@@ -0,0 +1,790 @@
...
+}
+
+/// The release directory name for a version and the digest of its binary.
+///
+/// The same `<tag>-<first twelve hex>` shape `install.sh` builds, so both
+/// routes name the same directory for the same released binary.
+#[must_use]
+pub fn release_name(version: &str, binary_digest: &str) -> String {
+    let short: String = binary_digest.chars().take(12).collect();
+    format!("v{version}-{short}")
```
mpyw: QUESTION: この short なんのために要るんだっけ？

mpyw:
フラットな関数多すぎて可読性終わってるので，構造体とメソッドにリモデリング中

rust/fslc/src/skills_install.rs
mpyw: 了解です。コードはそのままに、**なぜ今 Claude 決め打ちなのか**を doc コメントで残しました。リファクタで場所が動いて `skills.rs` の root にあります。

```rust
/// Claude Code only, for now. `README.md` documents the Claude Code skill
/// layout and no other, so a second built-in agent directory would be a
/// surface with nothing behind it. `--dir` reaches `.agents/skills`, or any
/// other location, in the meantime.
pub(crate) const AGENT_DIR: &str = ".claude";
```

「対応する余地はあるが今はやらない」を記録しておかないと、将来 `.agents` を見ない理由が分からなくなるためです。当面の逃げ道が `--dir` であることも併記しました。

rust/fslc/src/skills_install.rs
mpyw: 補足しました。`entries` と `links` それぞれに Key/Value と実例を書いています。

```rust
/// Key: the path relative to the skills directory. Value: the SHA-256 of
/// the bytes written there, as lower-case hex.
///
/// ```text
/// "fsl/SKILL.md" -> "b1946ac9...",
/// ```
```

`links` 側は `"fsl" -> "/home/me/.local/share/fsl/current/skills/fsl"` です。

ついでに設計判断も書いておくと、**この 2 つを別フィールドにしたのは検証方法が違うから**です。`entries` はファイル実体なので中身のダイジェストと突き合わせますが、`links` はシンボリックリンクなので「張り先が記録どおりか」しか見られません。1 つの map に混ぜると、値が digest なのかパスなのか型から分からなくなります。

rust/fslc/src/skills_install.rs
mpyw: 分かりにくかったです。書き直しました。

```rust
/// A file is at the path and no manifest of ours records it. Somebody
/// else's, sitting where a skill would go. Never written over, never
/// removed, only reported, because this tool did not put it there.
Foreign,
```

`Modified` との違いが肝で、そこを明示しました。

| 状態 | 誰のものか | install | uninstall |
|---|---|---|---|
| `Modified` | **うちのもの**を手で編集した | Blocked（`--force` で上書き） | Blocked（`--force` で削除） |
| `Foreign` | **他人のもの** | Skip して報告 | Skip |

`Modified` は「消していいか確認が要る」、`Foreign` は「そもそも触る権利がない」という区別です。`--force` が効くのが前者だけなのはそのためです。

`LinkState::Foreign` にも同じ説明を入れました。同じ語を同じ意味で使っている以上、片方だけ説明があるのは不親切なので。

rust/fslc/src/skills_install.rs
mpyw: `install.sh` との契約です。コメントを足しました。

```rust
// `install.sh:162` names the directory `$RELEASE_TAG-${CLI_HASH:0:12}`.
// That shell expansion takes twelve characters, so taking twelve here is
// what makes both routes name one directory for one released binary. The
// digest is in the name at all so two builds of one tag stay apart.
let short: String = binary_digest.chars().take(12).collect();
```

`install.sh:162` が `RELEASE_NAME="$RELEASE_TAG-${CLI_HASH:0:12}"` で作る名前と**同じものを `fslc` 側でも作る**必要があります。違う長さを取ると、同じバイナリに対して 2 つのディレクトリができてしまい、`install.sh` 経由と `fslc skills install -g` 経由で別の場所にペイロードが置かれます。

digest が名前に入っていること自体は、同一タグの別ビルドを分けるためです（`install.sh:166` が既存ディレクトリの中身を `diff -qr` で検証しているのと対になっています）。

実測でも一致を確認しました。

```
shasum -a 256 fslc | cut -c1-12  →  25984eeca29e
fslc skills install -g           →  v4.6.0-25984eeca29e
```

この指摘のおかげで「これはグローバルスコープ専用の知識だ」とはっきりしたので、リファクタでは `skills/global/location.rs` に置いています。以前は `locate.rs` でローカル解決と同居していました。

mpyw:
いったん形になってそうですが，もうちょいリファクタリング入れるかもしれないです

mpyw:
一応作ってみたものの、今のバージョニングと互換性を保ってスキルエンベッドするのかなりコストが高く感じました。結構やろうか迷うレベルになってきたなぁこれ…

- 今の `install.sh` がやってる互換性を捨てる
- https://github.com/mpyw/go-skill-embed みたいにスキルエンベッド部分だけを外出しする

やるとすればこのあたりか…

----

試作品↓

[mpyw/rust-skill-embed: Ship agent skills inside a Rust binary, and give that binary a skill install command](https://github.com/mpyw/rust-skill-embed)

mpyw:
mise 自体は asdf 互換の複数ディレクトリ構成の配布に対応してるっぽいから、無理にエンベッドしなくても今のままで案外いけたりするかな…？
抜本的に見直します

mpyw:
いったんクローズ

PR #1086, opened 2026-09-21
feat(release): publish a signed packslip so mise installs the Agent Skills with the commands

mpyw:
Resolves Phases 3, 4, and 5 of #1082.

Each release now carries a signed **packslip** manifest. mise reads it, installs
the commands from it, and fetches the Agent Skills declared beside them.

## Why

Only one install route ships the skills, and only for the machine it runs on.

| Route | Before | After |
| --- | --- | --- |
| `install.sh` | Linked, from a separate download | Unchanged |
| mise | **Nothing** | `mise skills sync`, per project, per pinned version |
| Single executable | **Nothing** | **Nothing**. See [What this does not do](#what-this-does-not-do) |

`AGENTS.md` requires a language feature to move with its skill references.
Nothing tied the shipped skills to the `fslc` that reads them. A mise user got
no skills at all.

mise gained a publisher-side mechanism for this in 2026.9.2, released
2026-09-07. This uses it rather than building another one.

## What changes

| | |
| --- | --- |
| `.github/workflows/release.yml` | Two `jdx/packslip@v1.2.0` steps and a gate over what they signed |
| `README.md` | A `packslip:` route beside the `github:` one, and which release each applies to |
| `install.sh` | One guard. It no longer takes a skill link another tool placed |
| `docs/RELEASE.md`, the release skill | The bundles join the asset inventory, and a mise install joins the verification |

No Rust source changes. The only Rust file touched is a test.

## The manifest

```yaml
- uses: jdx/packslip@74c75db7d611b36dd5a157999ec16cf823ccf80c # v1.2.0
  with:
    project: github.com/ymm-oss/fsl/fslc
    bin: fslc
    artifacts: >-
      release-assets/fslc-macos-arm64
      release-assets/fslc-linux-x64
      release-assets/fslc-linux-arm64
      release-assets/fslc-windows-x64.exe
    resources: ${{ steps.skill-resources.outputs.value }}
```

Two runs, one per command. The release publishes two commands as separate bare
executables, and one project cannot hold two `raw` artifacts for one platform.

Each asset is named in full. A `fslc-*` glob also matches every `fslc-lsp-*`
asset, which is the same trap the mise `matching` option sets and the reason
`README.md` uses an anchored `matching_regex`.

The skills are declared from the directory, never from a list kept beside it:

```
skill/fsl=repo:skills/fsl
skill/fsl-business=repo:skills/fsl-business
...
```

`repo:` pins each one to the release commit. #1084 removed three hardcoded
six-name lists because a skill added to `skills/` reached none of them. A fourth
list would have put it straight back.

> [!NOTE]
> The release asset layout does not change. packslip reads the bare executables
> as they are, and `fsl-skills.tar.gz` stays for `install.sh`.

## What is asserted rather than trusted

packslip reads the platform and the format off each file name. That is the whole
of what decides which binary a user is handed, so the workflow checks it against
an explicit table before the release is made public.

```
The fslc-lsp packslip does not describe the platforms these names claim.
```

The skill list is checked the same way, against `skills/` on disk.

The signing step runs after the remote asset check and before the release goes
public. The asset check compares the release against `release-assets/` exactly,
and the bundle does not come from that directory.

## The guard in install.sh

`install.sh` links `~/.claude/skills/fsl*` into its own versioned payload.
`mise skills sync --global` writes that same directory.

| | What it did to the other's link |
| --- | --- |
| mise | Refuses it, names it, and leaves it |
| `install.sh` | **Took it.** It moved anything in the way to `.pre-native-v3` |

A link this installer did not create now stays. A link whose target no longer
exists is still replaced, and a real directory is still backed up as before. The
rule names no other tool, because the reason is not about mise.

> [!IMPORTANT]
> This is the only collision, and it needs `--global`. Measured in an isolated
> `HOME`: `mise use -g`, `mise install` and `mise use -g` with
> `skills.auto_sync = true`, and plain `mise skills sync` all leave
> `~/.claude/skills` alone. mise links into the project by default.

## What this does not do

The single-executable download and the Windows route still get the commands and
no skills. Closing them means carrying the skills inside the binary.

That was built and measured on this branch before being dropped:

| | Bytes |
| --- | --- |
| `skills/` on disk | 263,362 |
| `fslc` without the payload | 37,026,304 |
| `fslc` with the payload | 37,790,288 |
| **Increase** | **763,984 (+2.02%)** |

It also added 22 entries to `rust/Cargo.lock`, of which 13 compile, and a second
major version of `syn`. #1082 records the full comparison.

## Evidence

Everything below was run against the real tools, not read about. Local versions:
mise 2026.9.5, packslip CLI 1.2.0.

| What | Result |
| --- | --- |
| `packslip create` over this release's asset names | `4 artifact(s), 8 resource(s), signed` |
| Platform inference from those names | `darwin/aarch64`, `linux/x86_64/gnu`, `linux/aarch64/gnu`, `windows/x86_64`, all `raw` |
| The workflow's gate, extracted and run against those bundles | Passes |
| The same gate against a bundle claiming `musl` | Refused, naming the command |
| What mise looks for | `packslip.fslc.sigstore.json`, which is what `create` wrote |
| `mise skills ls` / `sync` on a real packslip tool | Links made, handwritten skills untouched |
| `sync` over a directory or a link it did not make | `skipped ...: exists and is not a link mise made` |
| The `install.sh` guard, exercised from `install.sh` itself | Only another tool's link is left alone |

| Check | Result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | pass |
| `cargo test -p fslc-rust --test native_integration` | 6 pass |
| `cargo test --workspace --locked` | **still running.** This comment is updated when it lands |
| `tools/aggregate_changelog.sh check` | pass |
| `shellcheck -S warning install.sh` | pass |

Both new assertions carry a rejecting control. A `fslc-*` glob in place of the
four names fails with `packslip lost release-assets/fslc-macos-arm64`. The
installer guard removed fails with `the guard runs before any write`.

> [!WARNING]
> One thing cannot be verified before this merges. No release carries a packslip
> yet, so mise has never consumed this project's own. `docs/RELEASE.md` gains the
> step that checks it on the first release that does.

## Note on the stack

Built on #1083 and #1084. The branch is on a fork, so this targets `main` here
and its diff shows those commits until they merge.

---


rizumita (commented):
Thanks for this — building on mise's publisher-side packslip rather than inventing another mechanism is the right direction, and deriving the skill resources from `skills/` (instead of a fourth list) plus asserting the inferred platforms against an explicit table are both very nice. Uploading while the release is still a draft also fits our immutable-releases setting. A few things I'd like addressed before this leaves draft:

1. **Re-running the publish job after a partial failure (release.yml L411–471).** The action uploads the bundle itself (`upload: true` by default) before "Verify what the packslips claim" runs. If the fslc bundle lands and the lsp signing then fails (e.g. a transient Sigstore/Rekor error), or the gate fails, our release procedure says to re-run only the failed job — but the remote asset diff at L379–387 will then see the leftover `packslip.*.sigstore.json` and fail, so recovery needs manual asset deletion. Could you set `upload: false` on both steps, run the gate on the local bundles, upload both explicitly with `--clobber`, and then re-diff the remote assets against `release-assets/` plus the two bundles before making the release public? That also gives us a check that both bundles actually reached the release.

2. **The installer guard stops migrating legacy skill links.** Older installers linked `~/.claude/skills/fsl` to `$HOME/.fsl/skills/fsl` (or a checkout), and the current installer only prints a notice about `~/.fsl` without removing it. With `foreign_skill_link`, such a link still resolves, so it is now treated as foreign and left in place — the `.pre-native-v3` migration silently no longer happens, and the link breaks once the user removes `~/.fsl`. The same applies after changing `FSL_DATA_DIR`/`XDG_DATA_HOME`. The command-link path already recognizes legacy targets (install.sh L197/L242); could the skill path do the same (e.g. `$HOME/.fsl/skills/*`, `*/current/skills/*`), with a test for it?

3. **OIDC scope.** `id-token: write` / `attestations: write` are needed, but they now apply to the whole publish job, including `softprops/action-gh-release`. Would you consider moving signing into its own job (draft → sign → make public) so only the packslip steps can mint a token?

Questions:
- Does mise pin the full workflow identity (`.github/workflows/release.yml`) on first install, or only the `https://github.com/ymm-oss/fsl/` prefix? `pages.yml` in this repo also has `id-token: write`, so I'd like to understand what the signature rules out.
- For `repo:` skill resources, does the manifest carry a content digest, or is integrity only the commit pin?
- The lsp project is `github.com/ymm-oss/fsl/fsl-lsp` while the binary/assets are `fslc-lsp`. Since mise remembers the project/signer, is this name intentional before the first release fixes it?

Minor: the release skill's step 6 (what the tag push makes public) could mention the packslip bundles too, and the `${{ steps.*.outputs.bundle }}` expansions inside `run:` could go through `env:`.

This depends on #1083 and #1084; happy to re-review once those land and this is rebased (the #1084 commit here differs from that PR's current head).

